PR #3, opened 2023-06-14
imp simply translation

houdini91:
Simple translation of the current writer from onesbom to the upstream go-cyclonedx.

I basically did not yet clean it up , but i was hoping we can sync a bit see i am in the right direction. 

@puerco


pkg/writer/serializer_cdx14.go
```diff
@@ -182,43 +200,80 @@ func nodeToCDX14Component(n *sbom.Node) *cdx14.Component {
...
 
 	if n.Hashes != nil && len(n.Hashes) > 0 {
-		c.Hashes = []cdx14.Hash{}
+		c.Hashes = &[]cdx14.Hash{}
 		for algo, hash := range n.Hashes {
-			c.Hashes = append(c.Hashes, cdx14.Hash{
-				Algorithm: algo, // Fix to make it valid
-				Content:   hash,
+			*c.Hashes = append(*c.Hashes, cdx14.Hash{
+				Algorithm: NormalizeAlgo(algo), // Fix to make it valid
```
puerco, line 221: Awesome, another one you knocked off my todo-list :+1: 

I think we should abstract this away to share the algo normalization function betwe reader+writers plus other CDX versions?

pkg/writer/serializer_cdx14.go
```diff
@@ -138,25 +138,43 @@ func (s *SerializerCDX14) Serialize(opts options.Options, bom *sbom.Document) (i
...
+	}
+
+	if opts.Format.Type() != options.CDXFROMAT {
+		return fmt.Errorf("unsupported options, %v", opts)
+	}
+
+	encoder.SetPretty(true)
+
+	cdxVersion := cdx14.SpecVersion1_4
+	// 2DO - not sure if we should implicit switch case the suppported version or simply transform
```
puerco, line 169: Can you explain this a little bit more?

pkg/writer/serializer_cdx14.go
```diff
@@ -182,43 +200,80 @@ func nodeToCDX14Component(n *sbom.Node) *cdx14.Component {
...
 		}
 	}
 
 	return c
 }
+
+func NormalizeAlgo(algo string) cdx14.HashAlgorithm {
+	normAlgo := strings.ReplaceAll(strings.ToLower(algo), "_", "-")
+	switch normAlgo {
+	case "md5", "md-5":
```
puerco, line 251: I think we should adopt a single const for each of the algos in the format package. We should never get "sha1" or  "sha-1" in a protobom. The readers should normalize to predefined names

pkg/writer/serializer_cdx14.go
```diff
@@ -1,15 +1,16 @@
...
 	"fmt"
 	"io"
 	"strconv"
 	"strings"
 
 	"github.com/bom-squad/protobom/pkg/sbom"
 	"github.com/bom-squad/protobom/pkg/writer/options"
-	cdx14 "github.com/onesbom/onesbom/pkg/formats/cyclonedx/v14"
+
+	// cdx14 "github.com/onesbom/onesbom/pkg/formats/cyclonedx/v14"
```
puerco, line 12: Feel free tp completely drop this dependency :D :tada:

pkg/writer/serializer_cdx14.go
veramine: The .proto will have these kinds of fields, right?  So the reader will not really be able to choose what to call anything - the protobuf definition will have predefined field names and the readers will have to use those.

puerco: Ah you are right, we should enum them in the proto :+1:

houdini91: Added the enum, but it exposed something interesting (some algos are only supported in SPDX). 

```
enum HashAlgorithm {
    UNKNOWN = 0;
    MD5 = 1;
    SHA1 = 2;
    SHA256 = 3;
    SHA384 = 4;
    SHA512 = 5;
    SHA3_256 = 6;
    SHA3_384 = 7;
    SHA3_512 = 8;
    BLAKE2B_256 = 9;
    BLAKE2B_384 = 10;
    BLAKE2B_512 = 11;
    BLAKE3 = 12;

    //2DO what should we do
    MD2 = 13; // Only supported by SPDX
    ADLER32 = 14; // Only supported by SPDX
    MD4 = 15; // Only supported by SPDX
    MD6 = 16; // Only supported by SPDX
    SHA224 = 17; // Only supported by SPDX
  }
  ```

File: pkg/writer/serializer_cdx14.go (hunk 2 above)
Comment (puerco) [reply pkg/writer/serializer_cdx14.go:169]: Yeah I think the cdx 1.4 serializer should always return 1.4,  we can do a super simple writer for older versions that uses the same serializer and just have it return an older format.

File: pkg/writer/serializer_cdx14.go (hunk 3 above)
Comment (puerco) [reply pkg/writer/serializer_cdx14.go]: Yep, this will result in data loss, another one to document.

File: pkg/writer/serializer_cdx14.go [hunk 5]
```diff
@@ -182,43 +200,91 @@ func nodeToCDX14Component(n *sbom.Node) *cdx14.Component {
...
+				Type: cdx14.ExternalReferenceType(er.Type), // Fix to make it valid
 				URL:  er.Url,
 			})
 		}
 	}
 
 	return c
 }
+
+func NormalizeAlgo(algo string) cdx14.HashAlgorithm {
```
puerco, line 248: I'm going to merge this function but I think we should overload the proto with methods to output to each format. I'll do it in a follow-up to iterate faster, I'll tag you so that you can tell me what you think.

api/sbom.proto
```diff
@@ -153,3 +153,26 @@ message NodeList {
     repeated Edge edges = 2;
     repeated string root_elements = 3;
 }
+
+enum HashAlgorithm {
```
puerco, line 157: Cool, this was exactly the idea, just one detail: these should be used in the actual hash information (eg lines 30 & 126). I'll merge it as is and fix as it is non breaking and I'll fix it in a followup

puerco (approved):
This is great Mikey, thank you! I left a few comments about things we can fix outside of this PR.

PR #9, opened 2023-06-15
add cyclone to spdx analysis file

dn-scribe:
Added the cyclone-spdx-analysis.md file, which includes many of the translation issues.

### DIFF
--- a/docs/cyclone-spdx-analysis.md
+++ b/docs/cyclone-spdx-analysis.md
@@ -1,0 +1,386 @@
+# SBOM translation analysis
+
+## Background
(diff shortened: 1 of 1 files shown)


puerco:
Let's merge this Danny, I think the way to work through them would be to discuss them in separate issues like those you opened and then capture the resolution into doc, WDYT?

dn-scribe:
Sure.
I put it here for the benefit of whoever finds interest or may contribute. Please merge it.

jspeed-meyers (approved):
Merge this bad boy!

PR #12, opened 2023-06-18
Suggestion/cdx writer cleanup

houdini91:
I cleaned up serializer_cdx a bit, truthfully its for my own sanity but i am open to any comment suggestion or code standards we may want to impose.

I did not do much to the current flow but rather push in to related functions.

I not really sure if there was intention behind this but for every edge code seems to try and add all the components .
I think this causes some duplication
https://github.com/bom-squad/protobom/blob/main/pkg/writer/serializer_cdx.go#L136
(It felt wrong so i changed in PR but be glad to revert it if its actually needed).


jspeed-meyers:
> I not really sure if there was intention behind this but for every edge code seems to try and add all the components .
I think this causes some duplication
https://github.com/bom-squad/protobom/blob/main/pkg/writer/serializer_cdx.go#L136
(It felt wrong so i changed in PR but be glad to revert it if its actually needed).

@puerco?

pkg/writer/serializer_cdx.go
```diff
@@ -13,9 +13,19 @@ import (
 	"github.com/sirupsen/logrus"
 )
 
-type SerializerCDX struct{}
+type SerializerCDX struct {
+	rootDict       map[string]struct{}
+	addedDict      map[string]struct{}
+	componentsDict map[string]*cdx.Component
+	reflessList    []*cdx.Component
```
puerco, line 20: I think we can drop these in favor of a simpler serialization state object we can pass around to the internal methods. To simplify, we can drop:

- `rootDict` (see comment below)
- `reflessList` This one we should drop because nodes without an ID is something we cannot support (I'll detail more in a separate issue).

This means that we can have a single state object like this:
```golang
type cdxSerializationState struct {
	addedDict      map[string]struct{}
	componentsDict map[string]*cdx.Component
}
```

We could drop the added index too if we handle dependencies first. 

We can then pass around a pointer to the state to each internal function in a context or directly like this:

```go
  serializer.dependencies(state cdxSerializationState, *bom *sbom.Document)
```

The benefit of doing it like this is that we can handle as many serializations in parallel as we want without the need for mutexes, etc

pkg/writer/serializer_cdx.go
```diff
@@ -48,25 +94,28 @@ func (s *SerializerCDX) Serialize(opts options.Options, bom *sbom.Document) (int
...
 			for _, n := range bom.Nodes {
 				if n.Id == id {
-					rootComp := s.nodeToComponent(n)
-					doc.Metadata.Component = rootComp
-					addedDict[id] = struct{}{}
+					rootComp = s.nodeToComponent(n)
+					found = true
+					s.addedDict[id] = struct{}{}
 				}
 			}
```
puerco, line 120: Most of this logic can be removed in favor of the API functions I've started to build around the autogenerated objects, starting with https://github.com/bom-squad/protobom/pull/20 (see `document.GetRootNodes()` for a start)

puerco (commented):
Ah this is a great idea, we can split the logic into more focused functions to make them simpler. (Sorry I missed this PR when doing the reviews yesterday!)

I think this is a good idea, I have a few comments below :point_down:

puerco:
I opened https://github.com/bom-squad/protobom/issues/23 to expand the comment above on why there should never be Nodes without an Id (even when CycloneDX allows no bomRef).

puerco:
@houdini91 :
> I not really sure if there was intention behind this but for every edge code seems to try and add all the components .

Yeah, the demo serializer is probably doing all sorts of bad things, thanks for catching it. We should start cleaning it to make it "real" :)

pkg/writer/serializer_cdx.go
```diff
@@ -76,14 +125,24 @@ func (s *SerializerCDX) Serialize(opts options.Options, bom *sbom.Document) (int
 		}
 	}
 
-	// Next up. Let's navigate the SBOM graph and translate it to the CDX simpler
-	// tree or to the dependency graph
+	if !found {
+		return nil, fmt.Errorf("no root component")
```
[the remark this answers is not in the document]
houdini91: My bad, i was sure this is a required field in cdx, i also jumped to the conclution the issue is a result of a issue in the reader to the neutral document.

pkg/writer/serializer_cdx.go
```diff
@@ -31,15 +41,51 @@ func (s *SerializerCDX) Serialize(opts options.Options, bom *sbom.Document) (int
 	doc.Metadata = &metadata
 	doc.Components = &[]cdx.Component{}
 	doc.Dependencies = &[]cdx.Dependency{}
-	/*
-		if bom.Metadata.Date != nil {
-			doc.Metadata.Timestamp = bom.Metadata.Date.AsTime()
+
+	root, err := s.root(bom)
+	// 2DO Just warning until resolve BUG - the root spdx node is not found in the node list (search by ID)
```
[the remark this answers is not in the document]
houdini91: I think i am resolving this by removing the limitation entirely. 
```
root := s.root(bom) <- returns nil if no root
doc.Metadata.Component = root
```

What do you think?

pkg/writer/serializer_cdx.go
```diff
@@ -76,14 +125,24 @@ func (s *SerializerCDX) Serialize(opts options.Options, bom *sbom.Document) (int
...
+	if !found {
+		return nil, fmt.Errorf("no root component")
+	}
+
+	return rootComp, nil
+}
+
+// NOTE dependencies function modifies the components dictionary
+func (s *SerializerCDX) dependencies(bom *sbom.Document) ([]cdx.Dependency, error) {
+
```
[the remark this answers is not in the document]
houdini91: I was thinking about this, but was not sure how to separate it at the end i decided to skip the problem entirly.

IMHO there are two options
Options 1 - double iteration on dep:
In this option we reiterate on the dep once in the componentMap func and once in the dep func one, its slower but probably cleaner.
```
func (s *SerializerCDX) componentsMaps(bom *sbom.Document) ([]cdx.Dependency, error) {
    ...
    componentsDict := Map from s.nodeToComponent(n) ...
    
    componentsDictNew := componentsFromDependencies(bom, componentsDict) 
    
}

func (s *SerializerCDX) componentsFromDependencies(bom *sbom.Document, map[string]cdx.components)  (map[string]cdx.components, error)

func (s *SerializerCDX) dependencies(bom *sbom.Document) ([]cdx.Dependency, error) 
```


Options 2 - depHandlers:
IN this option basically we rename the dependencies func to something a bit more general and seperate the logic of each case in to a handler..
```
func (s *SerializerCDX) handleContains(dep cdx.Dependency, map[string]cdx.component ) ([]cdx.Dependency, error)

func (s *SerializerCDX) dependenciesWalk(bom *sbom.Document) ([]cdx.Dependency, error) {

switch ...
case "contains"
   s.handleContains()
case "depON"
   s.handleDepOn()
}

```


What do you think? how would you organise this

pkg/writer/serializer_cdx.go
houdini91: OK i added a comment to refactor the func when the PR is merged.
I also just noticing the rootDict is not used by anyone in the code so it can be simply removed for now anway

pkg/writer/serializer_cdx.go
houdini91: 1) Removing refless (placeholder for generating the ref for now)
2) removing rootDict (not used).

Q: Wouldn't two parallel serializations have two instances of the  SerializerCDX?

houdini91: I added `SerializerCDXState` as you suggested.
i was thinking maybe we like to use the context library that may be a good fit,
if you think its not needed i can remove this layer and simply provide the state pointer as you suggested.

houdini91:
@puerco Merged in main to unblock the conflicts.
I am not sure about the trick i do with the context , its sounds like we will want the library to include a context in at least the high level api we discussed,  not sure if its a overcomplication to propgate the state using the context object or maybe it is useful. 
What do you think?
In any case do you see any major issues in the cleanup PR?

puerco (approved):
OK, I brute forced the flaky test until it passed to merge this PR and I will fix it in another round of fixes.

PR #19, opened 2023-06-19
simple command line interface

houdini91:
I am adding a simple CLI interface to make things a bit more friendly..
This is a draft of something quiqly not trying to overstep the design phase of the CLI interface.

* Added Cobra root command calling the translation pkg.
* Added `output-format` `-o` flag.
* Skipping viper/configuration setting for now.

Run `go run cmd/demo-ingest/*go -h`
```
Translate SBOM formats

Usage:
  protobom [TARGET] [flags]

Examples:
  protobom [sbom-path] [flags]

	protobom [sbom-path] -o cyclonedx output a cyclonedx sbom
	protobom [sbom-path] -o spdx output a spdx sbom

Flags:
  -h, --help                   help for protobom
  -o, --output-format string   Select output format, [[text/spdx+text;version=2.3 text/spdx+json;version=2.3 text/spdx+text;version=2.2 text/spdx+json;version=2.2 application/vnd.cyclonedx+json;version=1.4 application/vnd.cyclonedx+json;version=1.5]] (default "application/vnd.cyclonedx+json;version=1.4")
  -v, --version                version for protobom
```


houdini91:
It needs a bit more work to get the format string more suitable to humans , sadly i am out of time today , continue tomarrow.

jspeed-meyers:
Thanks, @houdini91!

pkg/application.go
```diff
@@ -0,0 +1,53 @@
+package pkg
```
puerco, line 1: As these are the internals to power a demo app, can we make this package private?  Move it internal/pkg so that it cannot be imported outside of the modeule.

cmd/demo-ingest/root.go
```diff
@@ -0,0 +1,84 @@
...
+
+	"github.com/bom-squad/protobom/pkg"
+	"github.com/bom-squad/protobom/pkg/formats"
+	"github.com/sirupsen/logrus"
+	"github.com/spf13/cobra"
+)
+
+const (
+	PROTOBOM__LONG_DISCRIPTION = "Translate SBOM formats such as cyclonedx and spdx"
+	PROTOBOM_SHORT_DISCRIPTION = "Translate SBOM formats"
```
puerco, line 18:
```suggestion
	PROTOBOM_LONG_DESCRIPTION = "Translate SBOM formats such as cyclonedx and spdx"
	PROTOBOM_SHORT_DESCRIPTION = "Translate SBOM formats"
```

puerco (commented):
Nice @houdini91 !

This proves that. as the project matures, we need a stronger debugger tool, I have two comments, one is a simple typo but the other one is more important: let's be careful on what we export from this module.

Once you expose an API publicly, people will start using it and then it's hard to roll back. So let's start by hiding as much as we can.

pkg/writer/options/options.go
```diff
@@ -5,8 +5,8 @@ import (
 )
 
 type Options struct {
-	Format formats.Format
-	Indent int
+	Format formats.Format `yaml:"format,omitempty" json:"format,omitempty"`
+	Indent int            `yaml:"indent,omitempty" json:"indent,omitempty"`
```
puerco, line 9: I'm curious here, why the yaml annotations here? Any plans? :)

houdini91:
> Nice @houdini91 !
> 
> This proves that. as the project matures, we need a stronger debugger tool, I have two comments, one is a simple typo but the other one is more important: let's be careful on what we export from this module.

What do you mean debugger tool, like a sbom viewer?
I was thinking about how to bootstrap unittest and integration tests but i think i better start with writing up a suggestion issue for our CLI structure (user interface, build, test, release) so that we can discuss the details on it
 
> Once you expose an API publicly, people will start using it and then it's hard to roll back. So let's start by hiding as much as we can.
>As these are the internals to power a demo app, can we make this package private? Move it internal/pkg so that it cannot be imported outside of the modeule.

Can you clarify if your talking about the Translate func or Application Config structure or both?
I agree with you we should expose the minimum required to use the library.

I am ok with moving everything to an internal sub-directory but i just write down why i chose to expose this.
I think the highest level function used by our command is valuable for library users.
I also find it easier to explain the library when the command and library expose some reflection of each other.

```
import "github.com/bom-squad/protobom/pkg"

func main() {
  cfg := pkg.Application{}
  pkg.Translate(target, cfg)
}
```

FYI:
What i normally do is something weird like have the Application struct under `internal/config/application.go` and then in the `pkg/command.go`.
```
package pkg

type Application = internal.Application

Translate(cfg *Application)
```

pkg/writer/options/options.go
houdini91: i find yaml to be easier to use for a configuration file (comments, readability, templeting and so on..), i skipped this part in this PR because i was not sure we need a configuration file for this CLI.
I can remove it for now.

puerco:
> What do you mean debugger tool, like a sbom viewer?

No, not really, just a binary that let's us test everything without being the actual CLI we are building. Especially while we scaffold all the tests.

> I was thinking about how to bootstrap unittest and integration tests but i think i better start with writing up a suggestion issue for our CLI structure (user interface, build, test, release) so that we can discuss the details on it

Sounds good, it is a great idea to start working on. We should get a design doc/issue/etc going for the CLI tool to start thinking about how to expose the translation to the users.

> Can you clarify if your talking about the Translate func or Application Config structure or both?

I was talking about hiding all that's in `pkg/application.go`. Since this tiniy app is just a test binary (probably bound to be removed in the future altogether) we should hide it completely.

pkg/writer/options/options.go
puerco: Ah got it, yes I like the idea of a yaml config file too :+1:

houdini91:
I fixed the comments.

* Move application to internal subdir.
* Fixed typo

I also added some basic flags for humans but i was not sure if i am breaking @puerco vision of options based flags so put it in a seperate PR in my fork take a look and ping me if its ok to merge in to this PR as well or leave comments on it.

https://github.com/scribe-security/protobom/pull/1

Basically the new interface is something like 

```
key@mikey-Latitude-5520:~/scribe/DHS/protobom$ go run cmd/demo-ingest/*go examples/vt.spdx.json -f cyclonedx -v 1.3
Translate SBOM formats such as cyclonedx and spdx

Usage:
  protobom [TARGET] [flags]

Examples:
  protobom [sbom-path] [flags]

	protobom [sbom-path] -o cyclonedx output a cyclonedx sbom
	protobom [sbom-path] -o spdx output a spdx sbom

Flags:
  -h, --help                        help for protobom
  -f, --output-format string        Select output format type, [cyclonedx spdx] (default "cyclonedx")
  -m, --output-mime-format string   Select mime format, [json xml] (default "json")
  -v, --output-version string       Select version, map[cyclonedx:[1.3 1.4 1.5] spdx:[2.2 2.3]] (default "1.4")
      --version                     version for protobom
```

puerco:
I think this is what we need to start testing! My only request would be to just fold the CLI details into internal/ ([as mentioned here](https://github.com/bom-squad/protobom/pull/19#discussion_r1234699784))  to no export them and let's merge it :rocket:

houdini91:
@puerco Ok, i moved the application.go to internal.

Basica command

```
eitan@eitan:~/scribe/DHS/protobom$ go run cmd/demo-ingest/*go -h
Translate SBOM formats such as CycloneDX and SPDX

Usage:
  protobom [sbom] [flags]

Examples:
  protobom [sbom] -f cyclonedx                   output CycloneDX
  protobom [sbom] -f cyclonedx -v 1.3            output CycloneDX version 1.3
  protobom [sbom] -f spdx                        output SPDX
  protobom [sbom] -f spdx -v 2.3                 output SPDX SBOM version 2.3
  protobom [sbom] -f spdx -e xml                 output XML encoded SPDX

Flags:
  -e, --encoding string   Select encoding, Options=[json xml] (default "json")
  -f, --format string     Select format, Options=[cyclonedx spdx] (default "cyclonedx")
  -h, --help              help for protobom
  -v, --version string    Select version, Options=map[cyclonedx:[1.3 1.4 1.5] spdx:[2.2 2.3]] (default "1.4")
```

houdini91:
> Ah I see.. but IIUC this is just a test binary right? It is not supposed to be the final CLI binary. Or am I understanding it wrong? Tha's why I want to make it non importable.

I was thinking that the real CLI will be built ontop the POC but i think we need to sync on some details such as where is the proto/specific generated libraries/CLI hosted in relation to each other (not sure how this is normally done for protobuf based project? monorepo?)
I am totally ok with hiding everything for now, its easier to expose new interface in a library later in the process than hide a existing one.

puerco:
I think it is fine if we build a cli here in any dir (under another directory under `cmd/` or whatever) and its libs hidden under `/internal`. If it becomes something useful we can relocate it, if not we can use it as the foundation of a new design. My goal is to start using your CLI, it's looking awesome :)

houdini91:
@puerco Moved root.go to cmd/cli .
What about the changes to formats.go such as FormatOpt and the command line arguments lists, should we have something like `internal/cli/options.go` or something similar?

pkg/formats/formats.go
```diff
@@ -77,3 +99,35 @@ func (f *Format) Type() string {
 	}
 	return ""
 }
+
+type FormatOpt struct {
+	FormatType    string `yaml:"format-type,omitempty" json:"format-type,omitempty"`
+	FormatVersion string `yaml:"format-version,omitempty" json:"format-version,omitempty"`
+	MimeFormat    string `yaml:"mime-format,omitempty" json:"mime-format,omitempty"`
+}
```
puerco, line 107: Ah I see what you mean here.Well, we don'r really need these. The `format.Format` type can be overloaded to give you each of these parts. We just need to add the methods. For example:
```go
version := SPDX23JSON.Version()  // 2.3
standard := SPDX23JSON.Standard()  // spdx
encoding := SPDX23JSON.Encoding() // "json"
```
 
```

File: pkg/formats/formats.go [hunk 5]
```diff
@@ -77,3 +99,35 @@ func (f *Format) Type() string {
...
+
+func (f *FormatOpt) Select() (Format, error) {
```
puerco, line 109: What is the purpose of the Select() function? As it stands it just round trips and returns a format, why not just use the format.Format type directly? The whole formatOpt -> format round trip is just equivalent to just `var f format.Format("....")`. no?

puerco (commented):
OK, nice job tucking away the app internals Mikey, thanks!. I checked the rework and it feels that we are now getting a little bit of a blended mix of options everywhere. We should think of the options sets as belonging to each layer of the application stack so that things don't get mashed all over the place. When I say _stack_ I picture it something like this:

```
application layer
---
translator layer
---
reader | writer
---
serializers | parsers
---
protobom data layer
```

 These comments are focused on separating the user input flags (which are options at the application layer, ie coming from the cli or from a config file ) from the formats package (which does not really need options).

I tried to add verbose comments here, but let's discuss a plan to organize all the configuration option sets tomorrow when we meet.

pkg/formats/formats.go
houdini91: I guess its my bad naming of your toFormat

cmd/cli/root.go
```diff
@@ -0,0 +1,89 @@
...
+  {{.appName}} [sbom] -f spdx -v 2.3                 output SPDX SBOM version 2.3
+  {{.appName}} [sbom] -f spdx -e xml                 output XML encoded SPDX`
+)
+
+var (
+	version         = "0.0.0"
+	ApplicationName = "protobom"
+	Cfg             internal.Application
+)
+
```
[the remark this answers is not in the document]
houdini91: Agreed, 
i not sure what to do with the mimeUri Which is simply a reflection of the format field, i chose to include it as a option and autofill it when empty. What do you think.

Reading between the lines you don't think the validation of the formats we actuall support should be done in the cli layer but rather later on in the flow when we try to use the format instance?

houdini91:
@puerco Ping me if there anything missing from this PR

cmd/cli/root.go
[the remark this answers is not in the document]
puerco: We can if it shields from more expensive processing but we can worry about it once we start the 'real" CLI

puerco (approved):
Thanks @houdini91 this is great! Hopefully, this will give folks an easier way to start testing the underlying conversion libraries :)

PR #20, opened 2023-06-20
Initial Elements API Check-in

puerco:
This PR adds the first batch of API functions around the objects autogenerated from the proto.

All of these functions are not yet in use in the current reader/writer flows we are developing. The goal of this API is to make it easier to work with these objects from applications that use `protobom` as a library. In other words, this PR should not affect the translation logic we are building, it only supercharges the autogenerated objects to make them easier to use.

I wrote simple tests for most functions, I will add more test cases as the API moves forward.

### API Functions

The initial API functions include:

| Object | Method | Description |
| --- | --- | --- |
| Edge | Copy() | Deep copies an Edge object |
| Edge | PointsTo() | Returns a bool if an edge points to an element  |
| Node | Copy() | Deep copies a Node object |
| Node | Augment() | Agregates data to a node's fields from another node |
| Node | Update() | Updates a node's fields from another node |
| NodeList | Add() | Adds nodes from a second NodeList |
| NodeList | RemoveNodes() | Removes nodes from a NodeList |
| NodeList | GetEdgeByType() | Returns a pointer to an edge that matches From + Type  |
| NodeList | Intersect() | Returns a new NodeList combining the elements common to two NodeLists |
| NodeList | Union() | Returns a new NodeList combining all elements of two NodeLists |

Additionally the internal API adds the following functions to the NodeList object:

| Method | Description |
| --- | --- |
| indexNodes() | returns an index of all nodes, keyed by ID |
| indexEdges | returns and index of all edges, keyed by from and type | 
| indexRootElements | returns an index of all root elements | 
| cleanEdges | Cleans the edges of an element removing broken and empty edges, deduping them , etc |

### DIFF
--- a/pkg/sbom/edge.go
+++ b/pkg/sbom/edge.go
@@ -1,5 +1,25 @@
 package sbom
 
+// Copy returns a new edge with copies of all edges
+func (e *Edge) Copy() *Edge {
+	return &Edge{
+		Type: e.Type,
+		From: e.From,
+		To:   e.To,
+	}
+}
+
+// PointsTo returns true if an edge points to a node, in other words if it has
+// id in its list of Tos
+func (e *Edge) PointsTo(id string) bool {
+	for _, lid := range e.To {
+		if lid == id {
+			return true
+		}
+	}
+	return false
+}
+
 // ToSPDX2 converts the edge type to the corresponding SDPX2 label
 func (et Edge_Type) ToSPDX2() string {
 	switch et {
--- a/pkg/sbom/nodelist.go
+++ b/pkg/sbom/nodelist.go
@@ -18,18 +18,54 @@ func (nl *NodeList) indexIdentifiers() nodeIdList {
 // cleanEdges is a utility function that removes broken
 // connection and orphaned edges
 func (nl *NodeList) cleanEdges() {
+	// First copy the nodelist edges
+	newEdges := []*Edge{}
+
+	// Build a catalog of the elements ids
+	idDict := map[string]struct{}{}
+	for i := range nl.Nodes {
+		idDict[nl.Nodes[i].Id] = struct{}{}
+	}
+
+	// Now list all edges and rebuild the list
+	for _, edge := range nl.Edges {
+		newTos := []string{}
+		if _, ok := idDict[edge.From]; !ok {
+			continue
+		}
+
+		for _, s := range edge.To {
+			if _, ok := idDict[s]; ok {
+				newTos = append(newTos, s)
+			}
+		}
+
+		if len(newTos) == 0 {
+			continue
+		}
+
+		edge.To = newTos
+		newEdges = append(newEdges, edge)
+	}
+
+	nl.Edges = newEdges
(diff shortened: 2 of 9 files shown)


pkg/sbom/edge.go
```diff
@@ -1,5 +1,25 @@
...
+// Copy returns a new edge with copies of all edges
+func (e *Edge) Copy() *Edge {
+	return &Edge{
+		Type: e.Type,
+		From: e.From,
+		To:   e.To,
+	}
+}
+
+// PointsTo returns true if an esge points to a node, in other words if it has
```
jspeed-meyers, line 12:
```suggestion
// PointsTo returns true if an edge points to a node, in other words if it has
```

pkg/sbom/nodelist.go
```diff
@@ -0,0 +1,293 @@
...
+// nodeIndex is a dictionary of node pointers keyed by ID
+type nodeIndex map[string]*Node
+
+// edgeIndex is an index of edge pointers keyed by From elements and type
+type edgeIndex map[string]map[Edge_Type][]*Edge
+
+// rootElementsIndex is an index of the top levele elements by ID
+type rootElementsIndex map[string]struct{}
+
+// indexNodes returns an inverse dictionary with the IDs of thenodes
```
jspeed-meyers, line 15:
```suggestion
// indexNodes returns an inverse dictionary with the IDs of the nodes
```

pkg/sbom/nodelist.go
```diff
@@ -0,0 +1,293 @@
...
+	// Add a seen cache to dedupe edges when
+	// cleaning them up
+	seenCache := map[string]map[Edge_Type]*Edge{}
+	var seenEdge bool
+	// Now list all edges and rebuild the list
+	for _, edge := range nl.Edges {
+		newTos := []string{}
+		oldTos := []string{}
+
+		// Ig the from node is not in the index, skip it
```
jspeed-meyers, line 68:
```suggestion
		// If the from node is not in the index, skip it
```

pkg/sbom/nodelist.go
```diff
@@ -0,0 +1,293 @@
...
+			nl.Edges = append(nl.Edges, nl2.Edges[i])
+			continue
+		}
+
+		if _, ok := existingEdges[nl2.Edges[i].From][nl2.Edges[i].Type]; !ok {
+			nl.Edges = append(nl.Edges, nl2.Edges[i])
+			continue
+		}
+
+		// Add it her to the existing edge
```
jspeed-meyers, line 131:
```suggestion
		// Add it here to the existing edge
```

jspeed-meyers (approved):
Nice! The first tests in `protobom`!

Since this code isn't in any critical paths, I would say give it another 12 hours for any other reviewers to weigh in, but at that point, merge it and we can also add more or different tests later.

puerco:
OK, merging this, I'll start factoring out some of the logic in the reader/writer objects to use these new functions :rocket:

PR #22, opened 2023-06-21
Use NodeList in Document 

puerco:
This PR modifies the Document object to use a NodeList, allowing it to take full advantage of the NodeList API. An initial diagram is included to keep track of the current structure.


jspeed-meyers (approved):
Nice! LGTM.

PR #36, opened 2023-07-01
sniffer cleanup

houdini91:
@puerco 
New PR against, main stream your comments ;). 
Basic cleanup of the sniffer logic

Attaching @puerco Comment 
>  OK, the way the format detection works right now has two main goals:
> 
> Avoid reading the whole SBOM into memory
> It assumes the SBOM format can generally be found at the top of the file so it tries to optimize for less I/O by catching it early.
> The proposed patch breaks # 2 above. For example, if I point the sniffer it to the Kubernetes SBOM which is SPDX, it will read the whole 18 MB needlessly and read it again with the SPDX parser to find the format.
> 
> Do you want to rework the patch to maybe unify the reads and open the PR against main? There are some more comments about string conversion, etc but let's go through them in a PR to protobom/main


houdini91:
> The proposed patch breaks # 2 above. For example, if I point the sniffer it to the Kubernetes SBOM which is SPDX, it will read the whole 18 MB needlessly and read it again with the SPDX parser to find the format.

Can you elaborate ? (i loop over all sniffer for each fragment block).

pkg/formats/sniffer.go
```diff
@@ -25,72 +34,105 @@ func (fs *Sniffer) SniffReader(f io.ReadSeeker) (Format, error) {
...
-			break
+func (fs *Sniffer) sniff(data []byte) formatDetails {
+	for _, sig := range sniffFormats {
+		if format := sig.sniff(data); format.Type != "" {
+			return format
 		}
+	}
+
+	return formatDetails{}
+}
```
puerco, line 69: This is the function I was referring to. The previous algorithm captures less than one read the SBOM format. THe format gets detected in almost all cases by reading the first few lines.

In this new function, the format sniffer will reads the whole SBOM entirely for each failed format before doing now finally pass to find the version.

The example I noted was for the k8s SBOM. It will read the whole 18 MB file, detect it is not CycloneDX and then it needs to read it again from the beginning to find and return the SPDX version. And we're not taking into account XML or tag value yet.

pkg/formats/sniffer.go
```diff
@@ -25,72 +34,105 @@ func (fs *Sniffer) SniffReader(f io.ReadSeeker) (Format, error) {
...
-			strings.Contains(fileScanner.Text(), "'spdxVersion'") {
-			formatType = "text/spdx"
-			formatEncoding = JSON
-			if formatVersion != "" {
-				break
+type formatDetails struct {
+	Type     string
+	Version  string
+	Encoding string
+}
```
puerco, line 75: Why introduce another format to capture format info when we have this one? https://github.com/bom-squad/protobom/blob/d56806732603e3d147375f1f2fa33d9d0724ef8d/pkg/formats/formats.go#L10

puerco (commented):
> Can you elaborate i not sure i understand what needs fixing ? 

No problem @houdini91 , here are some comments. The idea of this library is to detect the SBOM format efficiently, that is with the least I/O possible on the SBOM file. This refactor would cause the format detector to iterate one time over all the SBOM data per format until one format matches. See comment below:

pkg/formats/sniffer.go
houdini91: Sorry for the missunderstanding
This function is called by parent on each file fragment.
The function sniff them recives the fragment in the bytes , and runs each sniffer only on the fragment them returns on the first found format so as the parent caller,  does come it actually going over the entire file?

pkg/formats/sniffer.go
houdini91: I agree I skipped that,  I was not sure how to set only parts of the format like the sniffer needs. 

Setting the type,version and encoded and seperatly, and I did not want to use 3 repetitive return values along the sniffer..

pkg/formats/sniffer.go
houdini91: Ok, i looked in to it, i think i now understand what your reffering to.

i had a bug i did not save the partial state of the format  (i was assuming the fragment is assumed to include all the details (type,name and version) for each format type.
I commits a fix for this specific bug (not all your comments resolved in this commit).
Does this resolve your concern about the IO reads?

I for some reason assumed the file fragment includes all the format details in one block (eg 512 byte block or something), i now tested it and found that it goes line by line and sometime the version is found in a separate line than the encoding.
I should never assume things sorry..

jspeed-meyers:
I'll take a review today or tomorrow. Thanks, @houdini91!

jspeed-meyers:
Hmm, @houdini91, would you be willing to show some basic stats on running this version of the sniffer versus the version currently in main?

It seems both you and @puerco agree on a key criteria by which to choose a sniffer:

> The idea of this library is to detect the SBOM format efficiently, that is with the least I/O possible on the SBOM file

But I can't tell just by looking at the code which version of the code is more efficient. If you selected at least two SBOMs (preferably several MBs), ran both, and then showed the timing stats for both, that would probably help resolve which sniffer approach is superior, at least on the efficiency metric.

How does that sound?

jspeed-meyers:
Thoughts from @veramine, @wey-chiang and others welcome!

veramine:
I'm not too worried about conversion throughput/speed/clock time  I would be in favor of a multiple-pass approach to improve accuracy.  I haven't looked at this code yet, but in general speed is not a factor for which to optimize.  Accuracy is far far more important.  Even if conversion takes multiple seconds, that would be ok for our particular use case.

jspeed-meyers:
> Accuracy is far far more important.

Question: are the approaches equally accurate? I currently don't know of evidence in any direction.

> Even if conversion takes multiple seconds, that would be ok for our particular use case.

I think for Chainguard's that this might not be the case. @puerco? My sense is that keeping the conversion very snappy is a technical priority for us.

houdini91:
I did not mean to modify the code flow, accurate or efficient. if i did it was a bug and hopefully i fixed it in the latest commit.
I also agree with @puerco approuch of iterating fragments of files between different sniffers, i only trying to clean up the responsibility of the sniffers to different formats so we have a clear view and interfaces both for our sake and library users.

@puerco @jspeed-meyers Take a peek see if it answers your concerns

jspeed-meyers:
Ahh, okay, @houdini91. I clearly mis-understood. Apologies!

I'lll defer to @puerco. Puerco, LMK if my review could be helpful.

puerco:
Ah now I see where the misunderstanding is:

>  i only trying to clean up the responsibility of the sniffers to different formats so we have a clear view and interfaces both

The problem with the approach right now is that is has to do one pass for each format. What we want is to detect the SBOM format in less than one read. Is we break the sniffers one per format, it will become more and more inefficient as the number of formats*encodings grow.

houdini91:
@puerco not sure I follow, if we want only one read then we need to probably exact the chunk size (line by line is not bug enough, also lines can be kinda logs of you don't indent .. )

Also do you mean my code in this PR breaks your original code efficiency? Can you try and point me to where you think it breaks. 
In the code I try to check cdx and spdx on each line read from file and if one of the sniffers find there respective formats the code sould break.

jspeed-meyers:
> The problem with the approach right now is that is has to do one pass for each format. 

@houdini91, are you saying the approach in this PR does NOT do one pass for each format? I suspect this is where the mismatch in understanding may lie.

If I'm correct, @houdini91, you're saying that this code checks line by line and on each line it checks for evidence of the document being SPDX or CycloneDX. The code therefore does not do a full ingestion of the document and check for SPDX and, if the document is not SPDX, then do another full ingestion of the document and check for CycloneDX. @houdini91, have I correctly represented the current functionality in this PR?

Another question to @houdini91: Why make this change at all? Is is better than the current approach in some way? Sorry, I think I've gotten a little lost.

@houdini91 and @puerco, we could also find a time to meet and discuss this in real time. Long GitHub issues can become confusing :(

pkg/formats/sniffer.go
```diff
@@ -25,72 +42,148 @@ func (fs *Sniffer) SniffReader(f io.ReadSeeker) (Format, error) {
...
 
-		// In JSON, the SPDX version field would be quoted
-		if strings.Contains(fileScanner.Text(), "\"spdxVersion\"") ||
-			strings.Contains(fileScanner.Text(), "'spdxVersion'") {
-			formatType = "text/spdx"
-			formatEncoding = JSON
-			if formatVersion != "" {
-				break
+func (c cdxSniff) Type() string {
+	return "cdx"
```
puerco, line 102: We should use the constants here,  but is this method needed?
```suggestion
	return CDXFORMAT
```

pkg/formats/sniffer.go
```diff
@@ -25,72 +42,148 @@ func (fs *Sniffer) SniffReader(f io.ReadSeeker) (Format, error) {
...
 		}
+	}
+
+	return state.Format()
+}
+
+type spdxSniff struct{}
+
+func (c spdxSniff) Type() string {
+	return "spdx"
```
puerco, line 134:
```suggestion
	return SPDXFORMAT
```

puerco (commented):
A few comments, as I mentioned in our sync I think we can simply the implementation by:

* Dropping the state struct in favor of a persistent state
* Dropping the context wired into the sniffer object
* Dropping the Type function and the use of the value in the sniffer interface as each sniffer is effectively getting a separate context.

puerco:
I synched with Mikey today about this PR and we're going to go ahead and split the detection into different objects working on data bits (as opposed to going through the full SBOMs data) as this PR is doing. After merging we'll optimize how we're cycling to lower the buffering cycles.

@houdini91 : It seems that the patch is failing to detect the CycloneDX example. As mentioned in the review comments above, I think we can simplify the implementation by dropping the state struct and keeping a persistent format. As these are internal details we can handle that a follow-up.

jspeed-meyers:
What's the status with this PR? Is further discussion needed? Some prototyping? Something else?

houdini91:
I think the pr is ready to be merged on the technical side after review of @puerco , aside of test not passing ofc. 
I did not debug the test issue deeply but hopefully i can resolve the issue end of next week (I am on vacation until then) so the pr can be merged.

houdini91:
@jspeed-meyers @puerco @manifestori 

I fixed the test,
I still have some linter issuer but it comes form a unrelated document , anyone can help me resolve this, so we can merge this

jspeed-meyers:
@houdini91:

I see this error:

```
Error: G601: Implicit memory aliasing in for loop. (gosec)
```

I would recommend running `gosec` locally to get more information, including the exact file and line that causes this error. That information should then lead you to a fix.

houdini91:
thanks,

> pkg/writer/serializer_cdx.go:184:20: G601: Implicit memory aliasing in for loop. (gosec)
					Dependencies: &e.To,

Its odd because it is not something i have changed in this PR @manifestori

houdini91:
@jspeed-meyers PR Is ready to me merged ..

jspeed-meyers:
Nice, @houdini91!

Let's get @puerco or @manifestori to take a look. But LGTM!

manifestori (commented):
@houdini91 Good work on capturing the nature of the internal implementation of Sniffer.

I think we should consider a mild refactor to `sniffFormat` / `Sniff` : 

My proposal would be: 

1) Remove sniffFormat interface, as it gives no real value, remember: Interfaces belongs to the Client, using them "for the sake of it" might lead to things like the `[]sniffFormat`.

2) I see why you needed to create multiple structs, you wanted them to implement the interface.  however, you can just take a more functional approach. 

Create `sniffCDX` and `sniffSPDX` functions. The both already capture their own `Type`, you can add a `SnifferFunc` interface if you want.

3) You can create a `SnifferFunc` slice like you did with with `[]sniffFormat` and range those. but you can simplify it and just call sniffCDX and sniffSPDX one after the other. I don't see how people add another format anytime soon.

This would reduce boilerplate, improve readability and make things easier to test :)

houdini91:
Ok, got it, i summarise in a issue so we can merge and continue this effort in another PR.
https://github.com/bom-squad/protobom/issues/74

houdini91:
@puerco I opened a new issue so we can track the pending issues from this PR.. (take a peek see if I missing something).
Imho they are not really blockers to merge this one which has been hanging for a while.

puerco:
OK, sounds good let's merge it as is and we can follow it up in subsequent PRs. Thanks Mikey! :+1:

puerco (approved):

PR #39, opened 2023-07-05
Add Equals Methods, deflake sbom.TestNodeListIntersect

puerco:
This PR adds an `Equal()` method to the `Edge`, `Node` and `NodeList` objects that lets users of the protobom API compare two objects to each other to find out if their data is the same:

```go
if node1.Equal(node2) {
   fmt.Println("Nodes are the same!")
}
```

It does so by adding an internal `flatString()` method to the objects that serializes their data into a comparable string.

As a drive-by, I've rewritten the the `sbom.TestNodeListIntersect` to use the new method from the nodelist effectively deflaking it.

### DIFF
--- a/pkg/sbom/edge.go
+++ b/pkg/sbom/edge.go
@@ -228,7 +228,7 @@ func (e *Edge) Equal(e2 *Edge) bool {
 }
 
 // flatString returns the edge serialized into a string that can be used
-// to indes or compare the contents of Edge e
+// to index or compare the contents of Edge e
 func (e *Edge) flatString() string {
 	return e.From + ":" + e.Type.String() + ":" + strings.Join(e.To, ":")
 }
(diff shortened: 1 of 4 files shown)


puerco:
CI is green :)

pkg/sbom/edge.go
```diff
@@ -218,3 +218,17 @@ func EdgeTypeFromSPDX2(spdx2Type string) Edge_Type {
...
+// Equal compares Edge e to e2 and returns true if they are the same
+func (e *Edge) Equal(e2 *Edge) bool {
+	if e2 == nil {
+		return false
+	}
+	return e.flatString() == e2.flatString()
+}
+
+// flatString returns the edge serialized into a string that can be used
+// to indes or compare the contents of Edge e
```
jspeed-meyers, line 231:
```suggestion
// to index or compare the contents of Edge e
```

jspeed-meyers (approved):
SWEET. Nice work!

jspeed-meyers:
cc @houdini91

houdini91 (approved):
Lgtm, thanks

PR #40, opened 2023-07-05
CDX Serializer cleanups

puerco:
This PR introduces two changes to protobom:

1. It refactors and renames some of the functions in the CDX serializer to hide them from the protobom API. This change changes the serializer to only expose the functions defined by Interface. Also some types and variables that are not needed outside the serializer are also hidden.
2. The second PR fixes the handling of nodes with autogenerated SBOMs. Our previous approach assumed that nodes without refs could find their way to the serializer. This is really no possible as all nodes are tracked in the graph using their IDs. The logic to autogenerate the node identifiers has been rewritten and has now moved to the `sbom` package. The new `sbom.NewNodeIdentifier` should be used to generate identifiers when ingesting CDX to properly tag those nodes who come from a refless component. From the commit message:

>  This commit refactors the handling of bom-refs in the CDX serializer.
>  The previous versions assumed that a protobom being serialized into CDX
>  could have refless nodes.
>  
>  This edge case is not possible, the whole protobom graph is keyed on the
>  node ids.
>  
>  What can happen is that a node may have been assigned an autogenerated ID
>  from a cdx component without a ref. When such a node is found, the serializer
>  will now remove the bom-ref to maintain the highest fidelity with the original
>  document.
    
/cc @houdini91


pkg/sbom/functions.go
```diff
@@ -1,10 +1,68 @@
...
 	cdx "github.com/CycloneDX/cyclonedx-go"
+	"github.com/google/uuid"
 )
 
+const NodeIdentifierPrefix = "protobom"
+
+var invalidIDCharsRe = regexp.MustCompile(`[^a-zA-Z0-9-.]+`)
+
+// NewNodeIdentifier returns an identifier string that can be used in a node
+// and that it is guaranteed to be compatible with CycloneDX and SPDX.
```
jspeed-meyers, line 19:
```suggestion
// and that is guaranteed to be compatible with CycloneDX and SPDX.
```

pkg/sbom/functions.go
```diff
@@ -1,10 +1,68 @@
...
+//
+// Without options, identifiers will be created using a new UUID and prefixed
+// with a prefix like "protobom-xx-yy--". This prefix allows serializers to read
+// characteristics of the identifier string, by looking for keywords before the
+// double dash. For example, "auto" means that it was autogenerated and did not
+// come from an ingested SBOM.
+//
+// Without any strings seeding it, NewNodeIdentifier generates the identifier
+// using an UUID. If a string is provided, it will be treated to remove any
+// invalid characters and used as the identifier.
```
jspeed-meyers, line 29:
```suggestion
// using an UUID. If a string is provided, any invalid characters will be removed and the 
// new string will be used as the identifier.
```

jspeed-meyers (approved):
NIce!

houdini91 (approved):
Awesome

PR #46, opened 2023-07-12
Protobom API Improvements

puerco:
This PR introduces two new functions to the protobom API:

- nodelist.RelateNodeListAtID(): his method takes a nodelist n2 and relates its top level elements to
the a node in nodelist nl1 using an edge type.
- nodelist.GetNodesByPurlType : return all methods that match a package URL type.

It also includes a final commit to further deflake some tests that deal with checking the consistency of the graph operations.

### DIFF
--- a/pkg/sbom/nodelist.go
+++ b/pkg/sbom/nodelist.go
@@ -4,6 +4,7 @@ import (
 	"fmt"
 	"reflect"
 	"sort"
+	"strings"
 
 	"github.com/sirupsen/logrus"
 )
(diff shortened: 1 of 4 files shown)


pkg/sbom/nodelist.go
```diff
@@ -419,9 +418,100 @@ func (nl *NodeList) Equal(nl2 *NodeList) bool {
...
+func (nl *NodeList) RelateNodeListAtID(nl2 *NodeList, nodeID string, edgeType Edge_Type) error {
+	// Check the node exists
+	nlIndex := nl.indexNodes()
+	nlEdges := nl.indexEdges()
+
+	if _, ok := nlIndex[nodeID]; !ok {
+		return fmt.Errorf("node with ID %s not found", nodeID)
+	}
+
+	// Check of we have edges matching
```
jspeed-meyers, line 441:
```suggestion
	// Check if we have edges matching
```

pkg/sbom/nodelist.go
```diff
@@ -419,9 +418,100 @@ func (nl *NodeList) Equal(nl2 *NodeList) bool {
...
+// GetNodesByPurlType returns a nodelist containing all nodes that match
+// a purl (package url) type. An empty purlType returns a blank nodelist
+func (nl *NodeList) GetNodesByPurlType(purlType string) *NodeList {
+	ret := &NodeList{}
+	if nl == nil {
+		return ret
+	}
+
+	for _, n := range nl.Nodes {
+		// I think the SPDX libraries have a bug where an extra slash is added when parsing purls
```
jspeed-meyers, line 480: Should we put in a PR to the upstream SPDX libraries? Feel free to do that as part of this work.

jspeed-meyers (approved):
Ship it! Sweet.

pkg/sbom/nodelist.go
puerco: I think I will, I think it will take more time to find it than the actual fix :D

PR #47, opened 2023-07-12
Add CycloneDX 1.4 unserializer

puerco:
This PR introduces the first iteration of the CycloneDX unserializer.

It supports most protobom fields (still missing are tools and persons) but it creates the correct protobom graph taking into account if the SBOM has a top-level component or not.

Requires the API improvements introduced in #46 (and probably a rebase)


jspeed-meyers (approved):
Cool. After rebasing, merge away, methinkgs.

PR #51, opened 2023-07-24
chore: devcontainer / precommit hooks

manifestori:
The objective of this PR is to simplify the procedure of bringing new developers onboard and ensure a smooth and consistent experience for them. Furthermore, the goal is to lay a solid groundwork for a successful open source project.

This includes the usage of [devcontainers](https://containers.dev/) which is already widely supported by many IDEs (even [Intellij](https://blog.jetbrains.com/idea/2023/06/intellij-idea-2023-2-eap-6/#SupportforDevContainers)! 

What does it do?
To simplify development, it installs a fully working developer environment that includes essential binaries such as go, python, and shell settings, all within a container. This environment can be tailored to our specific requirements.

It also allow users to run it via Github Codespaces (we can a get it for free as an OSS project)

Why? 
No longer need to maintain installation scripts, guides etc. 

What else? 
- I added `pre-commit` framework to start enforcing rules on commit.
- Some files were not correctly formatted 
- @manifestori (me) as a maintainer


jspeed-meyers:
@mikhailswift, @fkautz, or @veramine -- Any thoughts?

veramine:
I've never used devcontainers before either.  Would you please add a couple of sentences to the README describing how to take advantage of the idea?  It might even be nice to have a "GETTING STARTED" section in the README that starts with git pull or however the devcontainers work.  I'll test out the procedure if you want to write it!  Thx

puerco:
This LGTM, and I'm fine if we want to improve the documentation in a follow up. My only request @manifestori , can you please split the CODEOWNERS modification into its own commit? I think I would open a separate PR only for that change to keep the history isolated in case of a future audit?

manifestori:
Yes, for sure. Ill split it into a new PR. 

Also, will add a doc for "getting started" with devcontainer. Already have something in motion.

Thank you all 😊

go.mod
```diff
@@ -2,28 +2,24 @@ module github.com/bom-squad/protobom
...
 
-require google.golang.org/protobuf v1.31.0
-
 require (
 	github.com/CycloneDX/cyclonedx-go v0.7.1
 	github.com/google/uuid v1.3.0
 	github.com/sirupsen/logrus v1.9.3
 	github.com/spf13/cobra v1.7.0
 	github.com/stretchr/testify v1.8.4
+	google.golang.org/protobuf v1.31.0
```
puerco, line 11: The PR seems to be introducing lots of small white space and other minor changes. It is reshuffling `go.mod` here. Since it got into a merge conflict from the last dependabot update and it is not really related to PR's purpose I wonder if we should just drop this change from the commit? It's going to merge conflict #53 (or this one again) as it is.

manifestori: Yeah, I believe we can, I rebuilt go.mod since it got the common "problem" of multiple require () blocks. 

Also, I know it added a bunch of whitespaces, that's because I added a yaml linter. I will revert some of those and delay that into a later commit

puerco: Ah gotcha. OK, I pulled the require blocks together in go.mod, give it a rebase and you should see them joined now :+1:

puerco:
@manifestori: this should be good to go after the `CODEOWNERS` merge conflict is resolved :+1:

puerco (approved):

puerco (approved):
Thank you @manifestori !

PR #53, opened 2023-07-28
Fix flaking flat strings and add Checksums

puerco:
This PR fixes the flat strings used to serialize the various elements for comparison and checksumming. The first iteration was running into flakiness when some of the secondary structures were being serialized using their memory address. This PR fixes that and adds a lot more tests.

This PR is needed ahead of the conformance test framework which I'll PR once this fix merges.


jspeed-meyers:
@puerco -- Both the linter and the tests are failing. Any thoughts on what's going on? I'll review anyway.

pkg/sbom/node.go
```diff
@@ -211,14 +213,76 @@ func (n *Node) Equal(n2 *Node) bool {
...
+			for _, i := range n.Suppliers {
+				pairs = append(pairs, fmt.Sprintf("supplier:%s", i.flatString()))
+			}
+		case "bomsquad.protobom.Node.originators":
+			for _, i := range n.Originators {
+				pairs = append(pairs, fmt.Sprintf("originator:%s", i.flatString()))
+			}
+		case "bomsquad.protobom.Node.identifiers":
+			for _, i := range n.Identifiers {
+				pairs = append(pairs, i.flatString())
```
jspeed-meyers, line 231: Is the GitHub linter grumpy because the `flatString()` function is recursive?

pkg/sbom/node_test.go
```diff
@@ -184,3 +185,97 @@ func TestGetRootNodes(t *testing.T) {
 		require.Equal(t, tc.expected, nodes)
 	}
 }
+
+func TestNodeFlatString(t *testing.T) {
```
jspeed-meyers, line 189: Nice tests!

jspeed-meyers (commented):
Code looks good. But linter and tests are grumpy....

pkg/sbom/node.go
puerco: I had forgotten to check in a file. Fixed now!

jspeed-meyers (approved):
LGTM

pkg/sbom/identififer.go
```diff
@@ -0,0 +1,7 @@
+package sbom
```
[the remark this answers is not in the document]
puerco: Sorry, I need to work with my glasses on :upside_down_face: 
Done

Issue #56, opened 2023-07-31
Add Design Documents to Version Control


jspeed-meyers:
I think there are at least three design documents :)

puerco:
I think this was completed in https://github.com/bom-squad/protobom/pull/59 , feel free to reopen if not.

PR #57, opened 2023-08-02
Add PDF of SBOM format translation library proposal

jspeed-meyers:
Related to issue #56 

Outstanding questions: What format? PDF? Something else? PDF is convenient for representing the document without changing it. But it's bad for version control. Also, where to put this document within the file structure?

Alternatively, we could create a separate repo for design documents. Thoughts?


puerco:
I think we should distill this into a markdown file and store it in `docs/design/` or similar. I prefer if we can spare the download of the PDF to developers every time someone imports the library. We can still host the PDF on github (upload as if it was an image) and link it. Or in a gist, just not checked into git. WDYT?

veramine:
I agree

jspeed-meyers:
@puerco and @veramine: sounds good. @puerco, can you please then distill this into a markdown file and store it in `docs/design/`?

veramine:
I created a new pull request with a MD + the two images.

veramine (changes requested):
Agree we should use markdown, see new PR

jspeed-meyers:
Closing this PR. Thank you, @veramine!

PR #61, opened 2023-08-07
docs: add devcontainer getting started

manifestori:
A fundamental development getting-started doc. 

We cannot assume everybody uses VSCode so I made it more general, we can go back to it later.

Once we add `Make` commands we should update this docs as well.

### DIFF
--- a/docs/development.md
+++ b/docs/development.md
@@ -1,0 +1,21 @@
+# Devlopment
+
+### Development Container
+
+1. Docker / Podman
+2. VSCode (or other supported IDE)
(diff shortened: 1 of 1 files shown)


docs/development.md
```diff
@@ -0,0 +1,21 @@
...
+
+### Native
+
+1. Go 1.20+
+2. Make
+
+## Getting Started
+
+First, clone this repository and make sure you have your enviornment all setup. </br>
+For support IDEs users, you can simply run this project in a container (aka devcontainer): </br>
```
jspeed-meyers, line 17: I don't quite parse `For support IDEs users`. Maybe a slight rephrasing?

jspeed-meyers (approved):

PR #63, opened 2023-08-08
Drop demo CLIs from protobom repo

puerco:
This commit drops the previous CLI demos we had hosted in the protobom directory now that we have formalized the "official" CLI in bom-squad/bom-convert.

/cc @houdini91 @manifestori 

It also runs go mod tidy to drop unneeded dependencies.


jspeed-meyers (approved):
Kill it with fire 🔥

PR #65, opened 2023-08-09
Fix purl and CPE retrieval and storage

puerco:
This PR fixes a bug where all serializers and parsers were storing and reading the purl from the `node.ExternalReferences` list. The protobom model has a dedicated collection for software identifiers to keep them apart from the rest of the other references. This makes sure the data is being stored in the right place.


puerco:
Thanks @jspeed-meyers !
We have some tests to make sure ingestion and parsin works but the upcoming conformance tests will eventually catch this. Ingestion was working but it was recording the data in the wrong place, that is what the conformance tests will look for.

PR #66, opened 2023-08-09
nodelist.GetMatchingNode()

puerco:
This PR introduces a new method called `GetMatchingNode()` to the `NodeList` object. This method looks for a node that matches the passed node:

```golang
   matchingNode, err = nodelist.GetMatchingNode(testnode)
```

A "matching node" is a node that describes the same piece of software as the test node. This function is critical because node IDs are only valid within the document, so when looking for a matching node from a second SBOM, ID-based comparison cannot be done.

The function guarantees only one matching node. If there is ambiguity and two nodes from the nodelist fit the testnode, an error is returned.

The method looks for matching nodes by comparing the hashes first and if no match is done it tries to match based on the purl of the nodes. Purl matching needs a bit more polishing but it works for identical strings.

I wrote a lengthy test, it is included in the PR.

### DIFF
--- a/pkg/sbom/node.go
+++ b/pkg/sbom/node.go
@@ -299,3 +299,30 @@ func (n *Node) Purl() PackageURL {
 
 	return ""
 }
+
+// HashesMatch takes a map of hashes th and returns a boolean indicating
+// if the test hashes match those of the node. The algorithm will only take
+// into account algorithms that are common to the node and test set.
+//
+// In other words, if th has any hashes in algorithms without a peer in the
+// node, the function will ignore them and return true if others match,
+// silently ignoring those missing in the node.
+//
+// If either the node or the test hashes are empty, no match is assumed.
+func (n *Node) HashesMatch(th map[string]string) bool {
+	if len(n.Hashes) == 0 || len(th) == 0 {
+		return false
+	}
+	atLeastOneMatch := false
+	for algo, hashValue := range th {
+		if _, ok := n.Hashes[algo]; !ok {
+			continue
+		}
+
+		if n.Hashes[algo] != hashValue {
+			return false
+		}
+		atLeastOneMatch = true
+	}
+	return atLeastOneMatch
+}
--- a/pkg/sbom/nodelist.go
+++ b/pkg/sbom/nodelist.go
@@ -21,11 +21,13 @@ type edgeIndex map[string]map[Edge_Type][]*Edge
 // rootElementsIndex is an index of the top levele elements by ID
 type rootElementsIndex map[string]struct{}
 
-// hashIndex is a struct that indexes a node list by has values
+// hashIndex is a struct that indexes a NodeList by the hash values of its nodes
 type hashIndex map[string][]*Node
 
 // purlIndex captures the SBOM nodelist ordered by package url
 type purlIndex map[PackageURL][]*Node
+
+var ErrorMoreThanOneMatch = fmt.Errorf("More than one node matches")
 
 // indexNodes returns an inverse dictionary with the IDs of the nodes
 func (nl *NodeList) indexNodes() nodeIndex {
@@ -79,7 +81,7 @@ func (nl *NodeList) indexNodesByHash() hashIndex {
 }
 
 // Returns an indexed map of nodes by their package URLs. Note that more than
-// node may have the same purl.
+// one node may have the same purl.
 func (nl *NodeList) indexNodesByPurl() map[PackageURL][]*Node {
 	ret := map[PackageURL][]*Node{}
 	for _, n := range nl.Nodes {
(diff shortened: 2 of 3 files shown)


pkg/sbom/node.go
```diff
@@ -291,11 +291,36 @@ func (n *Node) Purl() PackageURL {
...
-			return PackageURL(e.Url)
+			return PackageURL(e.Value)
 		}
 	}
 
 	return ""
 }
+
+// HashesMatch takes a map of hashes and returns a boolean indicating
+// if the test hashes th match the nodes hashes. The algorithm will
```
jspeed-meyers, line 304:
```suggestion
// if the test hashes to match the nodes hashes. The algorithm will
```

pkg/sbom/node.go
```diff
@@ -291,11 +291,36 @@ func (n *Node) Purl() PackageURL {
...
+// only take into account hashes that are common in th and the nodes hashes.
```
jspeed-meyers, line 305:
```suggestion
// only take into account hashes that are common in the nodes hashes.
```

pkg/sbom/node.go
```diff
@@ -291,11 +291,36 @@ func (n *Node) Purl() PackageURL {
...
+// In other words, if th has hashes not found in the node, the function
+// wil return true, silently ignoring those that are not in the node.
```
jspeed-meyers, line 307:
```suggestion
// In other words, if there are hashes not found in the node, the function
// will return true, silently ignoring those that are not in the node.
```

pkg/sbom/node.go
jspeed-meyers: Ignore my suggested commits here and below if `th` is a variable reference. But perhaps consider another way to reference this variable. I was confused while reading this comment.

pkg/sbom/nodelist.go
```diff
@@ -21,6 +21,14 @@ type edgeIndex map[string]map[Edge_Type][]*Edge
 // rootElementsIndex is an index of the top levele elements by ID
 type rootElementsIndex map[string]struct{}
 
+// hashIndex is a struct that indexes a node list by has values
```
jspeed-meyers, line 24: What is "by has values?"

pkg/sbom/nodelist.go
```diff
@@ -56,6 +64,37 @@ func (nl *NodeList) indexRootElements() rootElementsIndex {
...
+			}
+			s := fmt.Sprintf("%s:%s", algo, hashVal)
+			ret[s] = append(ret[s], n)
+		}
+	}
+	return ret
+}
+
+// Returns an indexed map of nodes by their package URLs. Note that more than
+// node may have the same purl.
```
jspeed-meyers, line 84:
```suggestion
// one node may have the same purl.
```

pkg/sbom/nodelist.go
```diff
@@ -328,6 +367,87 @@ func (nl *NodeList) GetNodeByID(id string) *Node {
 	return nil
 }
 
+// GetMatchingNode looks up a node in the NodeList that matches the piece of
+// software described by testNode. It will not match on ID but rather matching
+// is performed by hash then by purl.
+//
+// This function is guaranteed to only return a node when there is a single node
+// matche. If more than one node matches after it returns ErrorMoreThanOneMatch.
```
jspeed-meyers, line 375:
```suggestion
// match. If more than one node matches after it returns ErrorMoreThanOneMatch.
```

pkg/sbom/nodelist.go
```diff
@@ -328,6 +367,87 @@ func (nl *NodeList) GetNodeByID(id string) *Node {
...
+		}
+	}
+
+	// Here, if we have exactly one node, then we have a match. If we have zero
+	// then we reindex and match on the purl. If more than one node matched on
+	// the hashes, we try to disabiguate by looking at the purl of the hash matches.
+	testPurl := node.Purl()
+	switch len(foundNodes) {
+	case 1:
+		// If  there is a single match, our job is done.
```
jspeed-meyers, line 409:
```suggestion
		// If there is a single match, our job is done.
```

pkg/sbom/nodelist.go
```diff
@@ -328,6 +367,87 @@ func (nl *NodeList) GetNodeByID(id string) *Node {
...
+		for _, n := range foundNodes {
+			return n, nil
+		}
+	case 0:
+		// No matches by hash, try to match by purl
+		// TODO(puerco): Purls should be normalized to match correctly,
+		// even more: ensuring correct globing of qaualifiers
```
jspeed-meyers, line 416:
```suggestion
		// even more: ensuring correct globing of qualifiers
```

pkg/sbom/nodelist_test.go
```diff
@@ -636,3 +636,325 @@ func TestEqual(t *testing.T) {
 		require.Equal(t, tc.shouldEq, res, msg)
 	}
 }
+
+func TestIndexByHash(t *testing.T) {
+	for label, tc := range map[string]struct {
+		sut            *NodeList
```
jspeed-meyers, line 642: What is "sut"?

pkg/sbom/nodelist_test.go
```diff
@@ -636,3 +636,325 @@ func TestEqual(t *testing.T) {
...
+						"sha1":   "7df059597099bb7dcf25d2a9aedfaf4465f72d8d",
+						"sha256": "c71d239df91726fc519c6eb72d318ec65820627232b2f796219e87dcf35d0ab4",
+					}},
+				},
+			},
+			expected:       hashIndex{},
+			expectedLength: 4,
+			mustEqual:      true,
+		},
+		"2 nodes, sahred hashes": {
```
jspeed-meyers, line 687:
```suggestion
		"2 nodes, shared hashes": {
```

jspeed-meyers (approved):
Just minor comment nits. This is cool base functionality!

pkg/sbom/nodelist_test.go
puerco: "Subject Under Test"

pkg/sbom/nodelist.go
puerco: Lol fixed it to "by the hash values of its nodes"

pkg/sbom/node.go
puerco: I reworded the whole comment because I could only read "puerco should sleep instead of writing this comment". I hope it is clearer now.

PR #67, opened 2023-08-09
Use ReadSeeker for reader.ParseStream.

wlynch:
Previously this was ReadSeekCloser, but the Close method was never used.

This also makes it easier to use with existing types like [bytes.Reader](https://pkg.go.dev/bytes#Reader), since you can now do:

```go
r := reader.New()
doc, err := r.ParseStream(bytes.NewReader(b))
```


jspeed-meyers (approved):
NICE!

puerco (approved):

Issue #64, opened 2023-08-09
Proposal: Change software identifier data type to map


puerco:
Right now, the software identifiers in protobom are stored in an array with the type and value each in its own field:

https://github.com/bom-squad/protobom/blob/a50c98ecf0438b3b899a3e780ec012f477ed4154/api/sbom.proto#L45
https://github.com/bom-squad/protobom/blob/a50c98ecf0438b3b899a3e780ec012f477ed4154/api/sbom.proto#L144-L147

This structure is designed in that way because documents in SPDX can have more than one identifier of any type (more than one purl, more than one cpe, etc).

This, however, poses a performance and usability problem with protobom. Finding a certain identifier is slow and cumbersome as we need to cycle all identifiers looking for a particular type. This is exponential when cycling through all packages in the sbom. The free string also poses a risk that keying on the right type may not always be accurate.

### Proposal:

I propose we change the identifier type to a map that uses a predefined map as the key:

```golang
 map[IdentifierType]string
```

This way we would get a one-shot access when trying to access the purl, etc:

```golang
purl := node.Identifiers[IdentifierType_PURL]
```

To keep with the lossless ingestion promise, we can create a new property "OtherIdentifiers" in the Node to capture when there are more than one identifier:

```golang
node.OtherIdentifiers = []struct {
  Type IdentifierType
  Value string
}
```

puerco:
Update: We discussed this change on the Aug 9th meeting and we're moving on to convert the identifier to a map.

puerco:
PR of implementation is in: https://github.com/bom-squad/protobom/pull/70

PR #70, opened 2023-08-09
proto: Change identifier field to map

puerco:
As discussed in the Aug 9th meeting, this PR modifies the proto data structure to transform the identifiers map to a new map for faster access and unambiguous keying.

The PR also updates all tests, serializers and parsers to use the new identifier data structure. The proto now includes a catalog enum of the types of identifiers we support.

Closes https://github.com/bom-squad/protobom/issues/64


jspeed-meyers (approved):
Wow, quick work @puerco! Nice

PR #81, opened 2023-09-01
Initial Conformance Framework Check-In

### LINKED ISSUE #24: Serializer Conformance Testing
## Problem:

We need to ensure the serializers create SBOMs consistently, with the expected degradation and structure.

## Proposal

We should create a collection of golden documents, each with different characteristics and challenges for the serializer. Some examples include: documents missing IDs, documents missing root nodes, documents with broken edges, etc.

Conformance tests would load the raw proto, feed the document into the serializer to generate a document in each format. If we have options for the serializers we should test all scenarios.

Paired with native tools to extract data from the documents (both SBOM tools and more lower level utilities like `jq` we then extract data from the documents:

image

Using this corpus of data, the generated documents and the tools we need, we create three kinds of tests:

1. **Reference Tests**<br>
These tests take the generated document and compare the data against the known features of the golden documents. This includes checking the number of packages, structure of the SBOM, edges, top level nodes, 

2. **Format Data Tests**<br>
These tests have to do with data specific to each format: Are the SPDX relationships sound? Is the CycloneDX dependency tree complete? Is the supplier data in the documents correct? Is the SPDX package verification code correct? etc.

3. **Format Conformance Tests**<br>
These tests mean running the "official" format validators against t the generated documents to ensure the serializer/writer pair is generating valid syntax.

## Open Questions

- Should we scaffold the tests early or wait until the serializers are more or less "complete"?

puerco:
This is the initial check-in of the conformance test framework. This is intended to be the bare-bones implementation
of the conformance framework to build on top of it. From the first commit description:

> This commit adds the initial conformance test skeleton.
> 
> At this time it is really simple, it has two simple tests using
> the built-in Equal() operator of the NodeList and a comparison of nodes.
>
> The idea of this check-in is to scaffold the way we run conformance tests,
> search for golden data, etc. This initial check of the conformance testing
> framework should let us start adding more conformance tests incrementally.

This PR enables the initial simple conformance tests and they will be run along the rest of the project's tests during the presubmits. It also adds some documentation and a utility to regenerate the golden samples. 

As mentioned above, this is the base to start adding more tests incrementally.

closes #24 


test/conformance/README.md
```diff
@@ -0,0 +1,81 @@
...
+The main goal of the protobom conformance testing framework is to ensure that
+each protobom serialization or unserialization result in equivalent documents
+every time.
+
+SBOM diffing is a complex problem. SBOMs cannot be compared as simple strings or
+comparing data structs. Verifying that documents are equivalent needs more precise,
+semantic comparison. 
+
+The approach the protrobom conformance tests take is to ask the same group of
+questions to a generated and to its equivalente golden sample and contrast the
```
jspeed-meyers, line 17: Typo in this line?

> The approach the protrobom conformance tests take is to ask the same group of
questions to a generated [SBOM?]

jspeed-meyers: `equivalente` -> `equivalent`

test/conformance/README.md
```diff
@@ -0,0 +1,81 @@
...
+answers. The more questions we ask, the stronger asurance the project can give
+of consistent parsing and serialization.
+
+Some sample questions that can be asked to documents:
```
jspeed-meyers, line 21:
```suggestion
Some sample questions that can be asked of documents:
```

test/conformance/README.md
```diff
@@ -0,0 +1,81 @@
...
+can be more than one sbom/proto sample per directory, in fact it is encouraged
+to add more data.
+
+### Generating the protrobuf Blobs
+
+When there is a structural change in protobom that results in expected changes to
+the parsed documents or serialized SBOMs, the protobuf blobs need to be regenerated.
+The conformance directory contains a generator utility in test/conformance/generator.
+
+Run the utility by running `make conformance`, it will cycle all the supported
```
jspeed-meyers, line 71:
```suggestion
Run the utility by running `make conformance`, it will cycle through all the supported
```

jspeed-meyers (approved):
SHIP IT. Like you point out, these can always be expanded.

test/conformance/README.md
puerco: fixed :+1:

Issue #89, opened 2023-09-02
Hash algorithm enumerations are not used


puerco:
## Short Description:

Nodes should use the enumerated hash algorithms.

## Details:

:warning: **Please note that the fix for this bug involves a breaking change** :warning: 

The proto definitions include an enumeration of the supported hash algorithms:

https://github.com/bom-squad/protobom/blob/4fc3791784be45015eb6a374db9d5972134f935f/api/sbom.proto#L153-L174

However, the node structure and the external reference is using a map of strings keyed with strings:

https://github.com/bom-squad/protobom/blob/4fc3791784be45015eb6a374db9d5972134f935f/api/sbom.proto#L28
https://github.com/bom-squad/protobom/blob/4fc3791784be45015eb6a374db9d5972134f935f/api/sbom.proto#L128

We should use the enumerated constants as keys in the hash entries. Having the value as strings opens up the use of protobom to all sorts of inconsistencies, case mixups and so on.

veramine:
My vote is to make the change.  I made this change in my local serializer_cdx.go but I'm embarrassed to check it in lol

```
< if algoVal, ok := sbom.HashAlgorithm_value[algoString]; ok {
-
> if algoVal, ok := sbom.HashAlgorithm_value[strings.ToUpper(algoString)]; ok {
```

puerco:
OK, I'm working on a PR to implement the change. I'll open it today and we can discuss it in the cohort meeting next Wednesday :+1:

veramine:
THANK YOU!

puerco:
Opened https://github.com/bom-squad/protobom/issues/89 to fix this issue

puerco:
Cohort emoji/vocally approved this change on the sep 6th meeting :+1:

PR #92, opened 2023-09-04
Added CDX serializer fields

veramine:
Added CDX serializer support for 

* Metadata.Tools, 
* Metadata.Name, 
* Component.Type (better validation from Node.PrimaryPurpose), 
* Component.Copyright, 

Also added a temporary strings.ToUpper() for HashAlgorithm_value lookup.

### DIFF
--- a/pkg/native/serializers/serializer_cdx.go
+++ b/pkg/native/serializers/serializer_cdx.go
@@ -273,8 +273,7 @@ func (s *SerializerCDX) nodeToComponent(n *sbom.Node) *cdx.Component {
 	case "":
 		// no node PrimaryPurpose set
 	default:
-		// Data loss here.
-		// TODO how do we handle when data loss occurs?
+		// TODO(degradation): Non-matching primary purpose to component type mapping
 	}
 
 	if n.Licenses != nil && len(n.Licenses) > 0 {
--- a/pkg/native/serializers/serializer_cdx.go
+++ b/pkg/native/serializers/serializer_cdx.go
@@ -275,7 +275,6 @@ func (s *SerializerCDX) nodeToComponent(n *sbom.Node) *cdx.Component {
 	default:
 		// Data loss here.
 		// TODO how do we handle when data loss occurs?
-
 	}
 
 	if n.Licenses != nil && len(n.Licenses) > 0 {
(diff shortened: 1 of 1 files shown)


veramine:
lol... always the whitespace...  let me try this again.

pkg/native/serializers/serializer_cdx.go
```diff
@@ -224,14 +240,42 @@ func (s *SerializerCDX) nodeToComponent(n *sbom.Node) *cdx.Component {
...
+		c.Type = cdx.ComponentTypeMachineLearningModel
+	case "operating-system":
+		c.Type = cdx.ComponentTypeOS
+	case "platform":
+		c.Type = cdx.ComponentTypePlatform
+	case "":
+		// no node PrimaryPurpose set
+	default:
+		// Data loss here.
+		// TODO how do we handle when data loss occurs?
```
puerco, line 277: We've been collecting instances of data loss with a comment like this:

```suggestion
		// TODO(degradation): Non-matching primary purpose to component type mapping
```

pkg/native/serializers/serializer_cdx.go
```diff
@@ -252,7 +296,7 @@ func (s *SerializerCDX) nodeToComponent(n *sbom.Node) *cdx.Component {
 	if n.Hashes != nil && len(n.Hashes) > 0 {
 		c.Hashes = &[]cdx.Hash{}
 		for algoString, hash := range n.Hashes {
-			if algoVal, ok := sbom.HashAlgorithm_value[algoString]; ok {
+			if algoVal, ok := sbom.HashAlgorithm_value[strings.ToUpper(algoString)]; ok {
```
puerco, line 298: OK, this could solve some cases but [cdx uses dashes in the algorithm labels](https://github.com/CycloneDX/specification/blob/cc15c8550e87dfa9ee051e6795e74a3c6c641264/schema/bom-1.5.schema.json#L705-L719) so, if the value comes from an [SPDX checksum](https://spdx.github.io/spdx-spec/v2.3/package-information/#710-package-checksum-field), they will not be valid.

Let's get it in for now, it will be solved once we resolve #89

puerco (approved):
LGTM, I just just a NIT suggestion for the data loss comment.

pkg/native/serializers/serializer_cdx.go
veramine: Yeah, #89 is definitely the better way.  This is a hack

puerco:
Awesome, thanks !

manifestori:
Well done @veramine !

jspeed-meyers (commented):
SWEET

PR #93, opened 2023-09-04
Use int-based hashmaps and enumerated algorithm values

### LINKED ISSUE #89: Hash algorithm enumerations are not used

puerco:
> **:warning: DO NOT MERGE YET :warning:** 
> To be discussed in the Sep 6th '23 cohort meeting

This PR changes protobom to use integer-based maps in the hash lists to prevent inconsistencies in the ingested and serialized algorithm names.  

See issue #89 for more details on the problem this PR fixes.

This is the full list of updates this PR addresses:

- [x] Update protobuf
- [x] Regenerate go types from proto
- [x] Update node and nodelists objects and methods (and their respective tests)
- [x] Update serializers
- [x] Update unserializers
- [x] Update new alpha SPDX3 code
- [x] Regenerate the existing conformance golden samples

Closes #89


jspeed-meyers:
> Aug 6th '23 

I think it's the Sep. 6th '23 meeting

veramine:
Would it make sense to use a HashAlgorithm as the key instead of an int32?  One is just a typedef to the other but calling it a HashAlgorithm may help humans if protoc will let you do it easily

puerco:
@veramine unfortunately you can't. protobuf3 does not allow enumerated values to be used as keys :/

veramine:
alright, no worries

veramine:
I'm in favor of this change

api/sbom.proto
veramine, line 1: It doesn't really matter that much but do you need to re-number "identifiers" here?  Were you just targeting hashes and accidentally changed identifiers as well?

puerco: Oops no sorry, I bumped the wrong field number. Good catch @veramine !
I've fixed it and repushed the branch.

veramine (approved):

puerco:
Cohort emoji/vocally approved this change on the sep 6th meeting :+1:

PR #94, opened 2023-09-05
Protobuf presubmit check

puerco:
This commit adds a new presubmit to check that the autogenerated code is up to date with the protobuf definitions. If the proto has been changed and the libraries need to be regenerated, this presubmit will fail.  The check script has pinned versions to ensure the check runs with the versions last used to generate the proto.

The hack README has the details on how to make the presubmit pass when it fails.

Note that the presubmit itself will fail until we merge https://github.com/bom-squad/protobom/pull/93 or regenerate the proto with the latest versions of `protoc` and `protoc-gen-go`. ([See a sample run here](https://github.com/puerco/protobom/actions/runs/6088891394/job/16520500043))

**Bonus:**  As a drive-by, I also added a bug fix to the fakes check to actually fail when it needs to.


puerco:
OK, now it matches the proto after #93 merged. I also fixed another bug in the bash and it is passing. PTAL :+1:

jspeed-meyers:
SWEET. Thank you, @puerco.

PR #95, opened 2023-09-07
ensure c.Type == file if node.Type is file

veramine:
My last change in this area was actually incorrect.  if protobom node.Type is "file" then CycloneDX Type MUST be "file".  

When ingesting an SBOM, this is what is supposed to happen:
* If we ingest cdx, the component.Type value should go to the primarypurpose field
** Additionally if component.Type is "file" the node should be marked as Node_FILE
* If we ingest spdx goes from primary purpose to primary purpose
When rendering the node to native SBOM:
* If it's SPDX the value goes from primary purpose to primary purpose
* If it's CDX:
** If the node is Node_FILE component.Type is file. Nothing else should happen or the graph will break.
** All others should go from the primary purpose to component.Type

Thanks @puerco for setting me straight here.  We would have had data loss without this change.

```
	if n.Type == sbom.Node_FILE {
		c.Type = cdx.ComponentTypeFile
	} else {
		switch strings.ToLower(n.PrimaryPurpose) {
		case "application":
			c.Type = cdx.ComponentTypeApplication
		case "container":
			c.Type = cdx.ComponentTypeContainer
		case "data":
			c.Type = cdx.ComponentTypeData
                ....
```

### DIFF
--- a/pkg/native/serializers/serializer_cdx.go
+++ b/pkg/native/serializers/serializer_cdx.go
@@ -279,7 +279,7 @@ func (s *SerializerCDX) nodeToComponent(n *sbom.Node) *cdx.Component {
 			// TODO(degradation): Non-matching primary purpose to component type mapping
 		}
 	}
-	
+
 	if n.Licenses != nil && len(n.Licenses) > 0 {
 		var licenseChoices []cdx.LicenseChoice
 		var licenses cdx.Licenses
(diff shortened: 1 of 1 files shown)


puerco (approved):
Nice! Thank you @veramine for getting this in before we cut the next release! :rocket:

PR #96, opened 2023-09-07
CDX Serializer: Add Primary Purpose to Type unit test

puerco:
This adds a test to ensure file nodes are rendered as file-typed components in CDX.

Follow up to https://github.com/bom-squad/protobom/pull/95

/cc @veramine 


veramine (approved):
yesssss LGTM

PR #97, opened 2023-09-07
Populate cdx Supplier from node.Suppliers

veramine:
Fill in CDX "Supplier" if at least one sbom Suppliers is set.  If more than one is set, we lose them.  I'm open to other ideas here but proposing this as a starting point.  

```
+       if n.Suppliers != nil && len(n.GetSuppliers()) > 0 {
+               // TODO(degradation): CDX type Component only supports one Supplier while protobom supports multiple
+
+               nodesupplier := n.GetSuppliers()[0]
+               oe := cdx.OrganizationalEntity{
+                       Name: nodesupplier.GetName(),
+               }
+               if nodesupplier.Contacts != nil {
+                       var contacts []cdx.OrganizationalContact
+                       for _, nodecontact := range nodesupplier.GetContacts() {
+                               newcontact := cdx.OrganizationalContact{
+                                       Name:  nodecontact.GetName(),
+                                       Email: nodecontact.GetEmail(),
+                                       Phone: nodecontact.GetPhone(),
+                               }
+                               contacts = append(contacts, newcontact)
+                       }
+                       oe.Contact = &contacts
+               }
+               c.Supplier = &oe
+       }
```

### DIFF
--- a/pkg/native/serializers/serializer_cdx.go
+++ b/pkg/native/serializers/serializer_cdx.go
@@ -339,6 +339,28 @@ func (s *SerializerCDX) nodeToComponent(n *sbom.Node) *cdx.Component {
 		}
 	}
 
+	if n.Suppliers != nil && len(n.GetSuppliers()) > 0 {
+		// TODO(degradation): CDX type Component only supports one Supplier while protobom supports multiple
+
+		nodesupplier := n.GetSuppliers()[0]
+		oe := cdx.OrganizationalEntity{
+			Name: nodesupplier.GetName(),
+		}
+		if nodesupplier.Contacts != nil {
+			var contacts []cdx.OrganizationalContact
+			for _, nodecontact := range nodesupplier.GetContacts() {
+				newcontact := cdx.OrganizationalContact{
+					Name:  nodecontact.GetName(),
+					Email: nodecontact.GetEmail(),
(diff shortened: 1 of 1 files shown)


pkg/native/serializers/serializer_cdx.go
```diff
@@ -339,6 +339,28 @@ func (s *SerializerCDX) nodeToComponent(n *sbom.Node) *cdx.Component {
 		}
 	}
 
+	if n.Suppliers != nil && len(n.GetSuppliers()) > 0 {
+		// TODO(degradation): CDX type Component only supports one Supplier while protobom supports multiple
+
+		nodesupplier := n.GetSuppliers()[0]
+		oe := cdx.OrganizationalEntity{
+			Name: nodesupplier.GetName(),
```
puerco, line 347: At some point perhaps we should think about how to translate here when dealing with organizations and persons. I think it is fine to add it like this now but the `sbom.Person` type no only is recursive, it can capture an organization (person.IsOrg) or individual. Perhaps there are better ways to match the cyclonedx field semantics?

puerco (approved):
Cool I agree, I think it is good enough to start. I think we should reflect on how to match the CDX structure  (see comment below) :point_down:

PR #98, opened 2023-09-08
chore: formatting and dependencies

manifestori:
This PR format some files that was never formatted, those should be auto formatted with pre-commit hooks. 

It also adds the correct version to .devcontainer of protoc and improves consistency across developers


jspeed-meyers (approved):
LGTM.

pkg/sbom/sbom.pb.go
```diff
@@ -1,7 +1,7 @@
 // Code generated by protoc-gen-go. DO NOT EDIT.
 // versions:
 // 	protoc-gen-go v1.31.0
-// 	protoc        v4.24.2
+// 	protoc        v4.24.0
```
puerco, line 4: the presubmit is failing because the proto is getting regenerated with a different (older) protoc:

```
-// 	protoc        v4.24.0
+// 	protoc        v4.24.2
```

puerco (commented):
Nice, just make sure that the protoc you are using matches the version of the generated code. See the comment below and [this document](https://github.com/bom-squad/protobom/tree/main/hack) for more info.

PR #99, opened 2023-09-11
initial version of spec doc

dn-scribe:
As decided in last SBOM Translator cohort meeting - here is an initial version of the doc.

### DIFF
--- a/docs/spec.md
+++ b/docs/spec.md
@@ -1,0 +1,80 @@
+# SBOM Translators Specification
+
+## Goals
+- Document decisions made by the cohort regarding the SBOM tranlation project.
+- Provide a concise reference for decisions - to be later translated into design, code and test cases.
+- Basis for future documentation - to allow users or the tools developed understand our translation decisions and rationale.
+
+Notes:
+-   Since most of the translation decisions are doen in the context of protobom readers and writers, this document is located here.
+-   This document is a documentation platform; the discussions are to be done through GitHub issues and the cohort meetings.
+
+# Design Decisions
+Note: this section documents main design decisions - these decisions do not directly map to spec, code or test cases but they do convay our approach and our understanding of the SBOM problem space.
+
+## A Translated SBOM is a New SBOM
+-   A translated SBOM is a new SBOM - it is not a copy of the original SBOM.
+-   As a result
+    -   The translated SBOM-ID should be generated upon translation and not copied from the original SBOM.
+    -   The creator of the SBOM should not be automatically translated from the original SBOM created.  
+
+## Conversion Tool Should Support a Configuration File to Enable Users to Control the Translation
+Uses:
+-   Override tools default behavior, in cases that the default behavior reflexes a specific translation decision.
+-   Enable users to add data that does not exist in the original SBOM but is required or at least supported by the target SBOM format.
+
+# Protobom Design Decisions
+## Protobom readers should not modify data
+
+
+# Common Subjects
+
+
+# CycloneDX <> SPDX Issues
+## Populating the SPDX [SPDXID](https://spdx.github.io/spdx-spec/v2.3/document-creation-information/#63-spdx-identifier-field) field
+-   The document ID will be randomized
+-   Enable user to override the ID through configuration.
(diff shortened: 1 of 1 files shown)


docs/spec.md
```diff
@@ -0,0 +1,80 @@
...
+- Document decisions made by the cohort regarding the SBOM tranlation project.
+- Provide a concise reference for decisions - to be later translated into design, code and test cases.
+- Basis for future documentation - to allow users or the tools developed understand our translation decisions and rationale.
+
+Notes:
+-   Since most of the translation decisions are doen in the context of protobom readers and writers, this document is located here.
+-   This document is a documentation platform; the discussions are to be done through GitHub issues and the cohort meetings.
+
+# Design Decisions
+Note: this section documents main design decisions - these decisions do not directly map to spec, code or test cases but they do convay our approach and our understanding of the SBOM problem space.
```
jspeed-meyers, line 13:
```suggestion
Note: this section documents main design decisions - these decisions do not directly map to spec, code or test cases but they do convey our approach and our understanding of the SBOM problem space.
```

jspeed-meyers (changes requested):
I'd personally rather have these decisions live in GitHub issues and the code. This document will, IMO, grow out of date too quickly and be too onerous to maintain.

I'd prefer to close this PR.

docs/spec.md
```diff
@@ -0,0 +1,80 @@
...
+Notes:
+-   Find the discussion [here](https://github.com/bom-squad/protobom/issues/8)
+
+
+# SPDX <> CycloneDX Issues
+## Singel Top Level Component in CycloneDX vs. Multiple Top Level Components in SPDX
+-   Generate a 'virtual' top level component in the translated CycloneDX SBOM.
+-   The virtual top level component details:
+    -   name - the (generated) SPDX DocumentName.
+    -   files analyzed should be false
```
puerco, line 76: In theory, we could set files analyzed to true and provide a package verification code.

docs/spec.md
```diff
@@ -0,0 +1,80 @@
...
+-   Enable user to override the name through configuration.
+
+Notes:
+-   Find the discussion [here](https://github.com/bom-squad/protobom/issues/6)
+
+
+
+## Populating the SPDX [Primary Package Purpose](https://spdx.github.io/spdx-spec/v2.3/package-information/#724-primary-package-purpose-field) field
+-   This field is to be populated from the CycloneDX component type field.
+-   If the file type of the cycloneDX component is ```file``` create an SPDX file component in the translated SBOM.
```
puerco, line 65: If we're writing a spec, there should not be specific tasks of going from CycloneDX to SPDX or viceversa. If we are specifying how protobom writes data, we should only talk about how data goes from protobom to the native formats.

This sentence, for example really should say that that `Node`s that have a type of `type_FILE` will be rendered as CyucloneDX components with a type of `file`.

puerco (commented):
@dn-scribe thanks for starting this doc danny. I think that what it is trying to achieve would be better served if we break the specification into format specific docs. 

All document serializations are one way, from protobom to native format. The caveats, hacks and tricks we need to document are specific to the way we sink data to each format.

At no point should talk about "conversion". If we need to, it should be in a doc in sbom-convert. I think we need to:

1. break the document, one for each native format
2. explain how data flows from a protobo, document to each sink, where each data bit lands
3. situations where data could be lost.

These documents are really the documentation of each serializer. In the future, these format-specific documents should also describe the options of the serializers (we're not there yet).

I added a few comments which should make it clearer 👇

docs/spec.md
dn-scribe: I did not understand your comment

docs/spec.md
dn-scribe: I understand the protobom-documentation approach, and I see how it serves the protobom use and development.

What I do not understand yet, is the sbom translator spec; I'm not sure it is reasonable to expect a user of the sbom translator cli to go to protobom documentation to understand the output. I also anticipate that most users will use the translator either as a tool or as a reference implementation without going deep into the code or the protobom documentation. 

How do you see this issue?

PR #102, opened 2023-09-12
Refactor Serializers / Unserializers 

manifestori:
In light of recent discussions, I have submitted this PR, This cleans up the `writer` package and `serializer` as well.

This also sets up the stage for future "SerializeOptions" allowing users to pass parameters to serializers and modify their behavior. 

This PR also adds CDX1.0-1.5.

Include writer test coverage (100%) and new fakes.

#72 - Closes some of the issues address here.


pkg/native/serializers/serializer_cdx.go
```diff
@@ -172,6 +185,7 @@ func (s *SerializerCDX) dependencies(ctx context.Context, bom *sbom.Document) ([
 	}
 
 	for _, e := range bom.NodeList.Edges {
+		e := e
```
manifestori, line 188: @houdini91 this is how u solve the "unsafe" memory in a loop.

pkg/writer/writer.go
```diff
@@ -1,56 +1,98 @@
...
-		Options: options.Default,
+var (
+	regMtx       sync.RWMutex
+	serializers  = make(map[formats.Format]native.Serializer)
+	defaultIdent = 4
+)
+
+func New(opts ...WriterOption) *Writer {
+	r := &Writer{
+		Indent: defaultIdent,
```
manifestori, line 28: Should we have a default format we want to write to?

manifestori: (I added one)

pkg/native/serializer.go
```diff
@@ -1,13 +1,21 @@
...
-	Render(options.Options, interface{}, io.Writer) error
+	Serialize(*sbom.Document, *SerializeOptions) (interface{}, error)
+	Render(interface{}, io.Writer, *RenderOptions) error
+}
+
+type SerializeOptions struct{}
+
+type RenderOptions struct {
+	Indent int
 }
```
manifestori, line 21: Those are redundant IMHO, and I think NewSerializer should include those options in instantiations (with or without options pattern), making Serializer simpler with only 

```go
type Serializer interface {
	Serialize(*sbom.Document) (interface{}, error)
	Render(interface{}, io.Writer) error
}
```

The only case against, is where you want to reconfigure a serializer forcing you to:
1) instantiate new serializer
2) register the new serializer with the writer

Still not sure the best route here as we don't "really" have options supported, planned.

pkg/native/serializers/beta/serializer_spdx3.go
```diff
@@ -102,9 +108,9 @@ type hashList struct {
 	HashValue string `json:"hashValue"`
 }
 
-type SPDX3 struct{}
+type SerializerSPDX3 struct{}
```
manifestori, line 111: @puerco I streamlined it to other serializer naming

adefee (approved):

pkg/native/serializer.go
puerco: I disagree with this idea on two fundamental grounds.

First, the serializers and unserializers should behave on each call according to the specifications sent by the caller. This means taking the execution specification (the options) on every run, they should be as stateless as possible. For example, if I have an application running, say a server responding to serialization calls, I want it to behave according to the parameters sent by the client. If the client needs to render using a set of options on one call and a different set of options on the next one, the application needs to react without having to destroy an reallocate the (un)serializer. Creating and destroying these drivers as a response to a change of the execution parameteres seems like an anti pattern.

Second, the `WithXXX()`/`WithoutXXX()` pattern: I can sympathize with the idea of preconfiguring the object at creation time. Unfortunately, the `With..` pattern does not fit this architecture. Options received by the library can be potentially piped down to the serializers and this makes it impossible to apply it to this scenario. The main problem is that you need to write a method to enable each option They may not be generalized, or they may even be unknown. Let me expand this idea with an example in the writer/serializer combo:

General options that apply to the writer or to all serializers can be implemented using `WithX...` , no problem. For example `WithIndent(4)`: Indentation level is an option that can apply to all json encodings and it is mostly harmless if it gets piped to serializers whose encodings don't have indentation (say SPDX tag-value). The problem, however, arises when considering options that apply to just one serializer, or (worse) when thinking about third-party serializers that get added as plug-ins. 

Consider an option that only applies to cyclonedx 1.4. Say you want an option to die when (cdx 1.5) Lifecycles are defined. In this case, you'd need an option `WithDieOnLifcycle()`. But this applies only to cdx 1.4. The method needs to exist, but it is not general for all serializers. Where do you define it? On the writer? On the general options? On the serializer?

And even worse, this option is probably something we will never have on the "stock" serializers. But suppose my employer wants to write a custom CDX serializer that has this option and use it as a plugin. Then the method needs to exist but there is no easy way for third party serializers to define these methods to make them available when you create the writer.

This is where empty interfaces (or generics, but let us not go into that just yet) come in handy. When passing an options struct on each run, the drivers can check an options set and check if a field is a known type and configure the run.

houdini91 (commented):
I agree some options can be given on the instance creation,
I don't have a strong opnion about this but let's consider If we expect the library user to use a single seralizer instance multiple times with different options.

pkg/native/serializer.go
houdini91: I agree some options can be given on the instance creation,
I don't have a strong opnion about this but let's consider If we expect the library user to use a single seralizer instance multiple times with different options.

pkg/native/serializers/beta/serializer_spdx3.go
puerco: Lol, I was actually trying to start destuddering all the serializers and unserializers. Since the serializers now live in a `serializers` packag, I thought it would be better to have:

`serializers.SPDX3` instead of `serializers.SerializerSPDX3`.

I don't care a lot about names and we can change them if they feel more natural.

pkg/native/serializers/serializer_cdx.go
houdini91: Thanks, I am interested can you elaborate on way this is safer?

pkg/writer/implementation.go
puerco, line 1: Why are you removing the implementation here? By moving the functionality one level up (from the implementation to the actual writer object) you are losing the integration simulation of each of the writer functions steps. In other words you are making tests less granular and the functions become a black box where you only can observe them from the outside.

For example, at present, there are test cases where we ensure we have a controlled behavior when `implementation.GetFormatSerializer()` fails or succeeds. In future improvements to the tests we can ensure it returns expected values when fed certain arguments, etc. In the rewrite you are proposing, the step functions of `WriteStream()` and `WriteFile()` are now hidden away in the black box and we can't test what happens when they succeed, fail or return unexpected data.

While the serializer itself is now mockable (which is good) the writer internals are no longer so (which is very bad).

puerco (commented):
Thanks for the PR @manifestori. Thanks for adding to the project, there are some concerns in the PR which I think we need to clear.

Frist. please refrain from breaking the API in the PR unexpectedly (see the comments below). While we still can change the API, we should do so carefully as some of us already have projects that depend on protobom so we need to tread carefully. Especially when the benefits of the changes are limited.

I have some concerns about changing the architecture of some of these objects, I'm referring to the away we wire the options and break the implementation pattern. I added detailed comments on where I see some problem areas with the proposed changes.

Finally, as we discussed when we met, try to scope the pull requests to atomic changes. I feel bad having to suggest changes on such a big effort and these PRs take a loooot longer to review. This PR is injecting 5-6 considerable changes in one and while some of them could go in as they are, they are all held back by the problematic areas of the PR.

pkg/native/serializer.go
puerco: Right, this is the same take I have on this (see the comment on my review).

pkg/writer/writer.go
puerco: Could be, but the question is reacting to a problem. Right now the format is an option in the writer but as we've been seeing when using the library in production, I think this was a design flaw. My feeling now is that you should be able to command, on each run, which format to use when writing the documents. Something equivalent to "_write this doc in format x to this stream_", like this:

`writer.WriteStream(doc, format, r io.Writer)`

Unfortunately, this would break the API. To bridge it we could still implement a `write.WriteFormatToStream()` or something

pkg/native/serializers/serializer_cdx.go
manifestori: In each iteration, the iteration variable is assigned the value of the next element in the range expression. The variable itself doesn't change, only its value does. That's why the expression 'e' is pointing to the same location in memory throughout the loop. (this is why its unsafe)

This is why using 
`e := e` fixes this, because you create a new `e` variable in scope.

pkg/writer/writer.go
manifestori: (Good morning @puerco :) ) 

I feel perfectly fine with breaking this now, I agree with your suggestion, makes a lot of sense to avoid creating multiple instances of a writer. also if someone would want to wrap this with "helpers" and implement 
```go
w := writerhelper.wrap(writer.New())
w.WriteCDX15Stream(doc, wr)
...

it wouldn't be possible without allowing someone to change the format you serialize to dynamically

File: pkg/native/serializers/beta/serializer_spdx3.go (hunk 4 above)
Comment (manifestori) [reply pkg/native/serializers/beta/serializer_spdx3.go:111]: I can do us a favor and rename all existing Serializers to shorted names. 
I prefer the shorter version, just thought we should be consistent

Comment (manifestori) [reply pkg/native/serializers/beta/serializer_spdx3.go:111]: serializers.SPDX3
serializers.SPDX23
serializers.CDX
...

File: pkg/writer/implementation.go
Comment (manifestori) [reply pkg/writer/implementation.go:1]: @puerco black box implementations are good, this is merely me following open-close principles. Testing private functions should be done via public API, and with 100% coverage of the writer packages (really, 100% coverage).

I agree that since we use a global serializers registry, it would be smart to change `func (w *Writer) getFormatSerializer` to just `func GetFormatSerializer`. however, this is not true for other "helper functions" we had in `implementation` (OpenFile for example). 

In the past, I used to wonder why some packages don't have an outward-facing interface. The answer was always "interface pollution." If someone wants to mock the writer, it can be done easily. but its the "user" responsibility.
For example:

```go

// not mockable
func write(w *protobom.Writer) {
	w.WriteStream(nil, nil)
}

func main() {
	w := writer.New()
	write(w)
}

// mockable
type BOMWriter interface {
	WriteStream(bom *sbom.Document, wr io.WriteCloser) error
	WriteFile(bom *sbom.Document, path string) error
}

// not mockable
func write(w BOMWriter) {
	w.WriteStream(nil, nil)
}

func main() {
	w := writer.New()
	write(w)
}

func test() {
	w := writermock.New()
	write(w)
}
```

Also note that since package level functions cannot be mocked (Register/Get Serializers), the "mockability" can be achieved by making a new struct "SerializerRegistry" and having it as a dependency for Writer (I don't think it necessary) that would make a "concise" implementation
In the existing implementation, you cannot really swap the implementation either.

manifestori: Also, please note that since "impl" is an instance, if someone does reimplement it, we would have to overhaul the entire writer package anyway to include a "options pattern" (or we break it with impl as parameter).

manifestori: https://github.com/bom-squad/protobom/pull/102/files#diff-3fc0a789ab29ceb166e8753d39a23b6c660c7e5ab41c5ba2076663f64ccc9b09R68

updated version, also added unregister function

pkg/native/serializers/beta/serializer_spdx3.go
manifestori: Done :)

pkg/native/serializer.go
```diff
@@ -1,13 +1,21 @@
...
+
 import (
 	"io"
 
 	"github.com/bom-squad/protobom/pkg/sbom"
-	"github.com/bom-squad/protobom/pkg/writer/options"
 )
 
+//counterfeiter:generate . Serializer
 type Serializer interface {
```
manifestori, line 12: @puerco @houdini91 in light of our discussion regarding interface pollution, I honestly think this file/interface belongs in either "writer" (the "user" of the serializer) or in a "domain" package. 

Serializers could implement the interface with ref to writers package.

`var _ native.Serializer = &CDX{}`
to 
`var _ writer.Serializer = &CDX{}`
making the beta clearer
`var _ writer.Serializer = &SPDX3{}`

Interface pollution - users is responsible for its interface, the implementation doesn't care (but it sometimes should, this is why we verify it does :P )

pkg/native/serializer.go
```diff
@@ -1,13 +1,21 @@
...
-	Serialize(options.Options, *sbom.Document) (interface{}, error)
-	Render(options.Options, interface{}, io.Writer) error
+	Serialize(*sbom.Document, *SerializeOptions) (interface{}, error)
```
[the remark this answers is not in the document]
manifestori: I know this is breaking, as this PR has some breaking changes that aim to improve the outward facing api.

I've never encountered a function with "options" as first arg, before the "subject" (doc)
also, I think we still have a debate regarding the options for both functions, would be easy to remove (or change) later or add 

`SerializeWithOptions(doc, opts)` 
and keep `Serialize(doc)` cleaner

pkg/writer/implementation.go
manifestori: (btw: I also plan to do the same design change) to `Unserializer` (also in the works 🏗️ )

pkg/native/serializer.go
manifestori: It would be more logical to relocate the Serializer interface to the writer package. 
I just don't want to make the writer/native/serializer tightly coupled

manifestori: Also, I want to use a DTO if we pass "config" to the writer. I think serializer should not accept the writer config (cyclic dependency) but the writer should create the Render/SerializeOptions from the config.

manifestori: https://github.com/bom-squad/protobom/compare/refactor-core-serializers...demo-writer-with-config?expand=1

I pushed this new branch to showcase the general idea. `demo-writer-with-config`

what changed:
Writer now has default config (on instantiation)
Writer now have new functions `WriteStreamWithConfig` and `WriteFileWithConfig` to allow setting those configs on the fly.
Writer existing function uses default config
This config is converted to Render/SerializeOptions using DTOs (writer.Config -> native.RenderOption)

manifestori: There is still an issue with passing arbitrary options to serializers, but this is OOS in this PR.

manifestori: > if want us to pass arbitrary options to known/unknown parsers, option like `DieOnLifcycle` (CDX parser option) We need to either add "DieOnLifecycle" to "SerializeOptions" struct, make it an interface{} or god forbid use generics 😂
> 

Regarding this one, so if you think the best way to support an unknown prop bag for a serializer is to pass as an interface{} , that's fine by me. but the writer would have to accept a Config with 

```go
SerializeOptions interface{}
RenderOptions interface{}
```
and the writer will delegate those to their respective function calls.

manifestori:
> Thanks for the PR @manifestori. Thanks for adding to the project, there are some concerns in the PR which I think we need to clear.
> 
> Frist. please refrain from breaking the API in the PR unexpectedly (see the comments below). While we still can change the API, we should do so carefully as some of us already have projects that depend on protobom so we need to tread carefully. Especially when the benefits of the changes are limited.
> 
> I have some concerns about changing the architecture of some of these objects, I'm referring to the away we wire the options and break the implementation pattern. I added detailed comments on where I see some problem areas with the proposed changes.
> 
> Finally, as we discussed when we met, try to scope the pull requests to atomic changes. I feel bad having to suggest changes on such a big effort and these PRs take a loooot longer to review. This PR is injecting 5-6 considerable changes in one and while some of them could go in as they are, they are all held back by the problematic areas of the PR.

Thank you for your comments, I tried to scope this to a very limited set of changes. 
I'm fine with reverting the changes to `Serializer` interface as it's the mildest one and note the core of the PR. we need to resolve options issue before we can merge this IMO

manifestori:
I suggest we defer the discussion for "Options" to this PR -> 

https://github.com/bom-squad/protobom/pull/104

I can wait with this one (if we have no other pending issues) until the work on #104 is done and agreed.

pkg/writer/writer.go
veramine: If ever there were a time to break the API, now is the time!  My suggestion would be to make it right now for the long term, and not worry so much about breaking the API in the immediate term before we have significant usage.

manifestori:
@puerco @veramine @houdini91 

I just merged the unserializer/reader, can we get this through the finish line? any final changes/amendments you had in mind?

houdini91:
Lgtm great work

houdini91 (approved):

manifestori:
Can I get this approved @veramine @jspeed-meyers @bardenstein I need a "Codeowner" approval.

As @houdini91 and I agreed, all open issues could be moved into a new PR/Issue. Lets get this merged and move forward with and following outstanding discussions.

veramine:
This is all ok with me but I was waiting to see if more discussion about Options was going to happen.  Let's discuss on the Sept 4 cohort call and then finalize after the meeting.  Thanks!

adefee (approved):

puerco (approved):
Thanks @manifestor awesome worki!!

puerco (approved):
Thanks @manifestor awesome worki!!

veramine (approved):
This is great.  As discussed today, I would prefer the default CDX serialization format to be 1.4.  It's fine with me to change that subsequently though.

jspeed-meyers:
Let's merge this then, right?

veramine:
We talked in the meeting this morning about @manifestori merging tomorrow morning.  @puerco was going to make a couple more comments today.  We are also working toward an imminent 0.2 release.

pkg/native/serializers/serializer_cdx.go
```diff
@@ -368,17 +423,26 @@ func (s *SerializerCDX) nodeToComponent(n *sbom.Node) *cdx.Component {
...
+	encoding, err := cdxformats.ParseEncoding(s.encoding)
+	if err != nil {
+		return fmt.Errorf("getting CDX encoding: %w", err)
+	}
+
+	encoder := cdx.NewBOMEncoder(wr, encoding)
 	encoder.SetPretty(true)
 
-	if err := encoder.EncodeVersion(doc.(*cdx.BOM), cdxVersion); err != nil {
+	if err := encoder.EncodeVersion(doc.(*cdx.BOM), version); err != nil {
```
puerco, line 445: We need to test if this interface can be cast, otherwise, we could introduce a panic here:
```suggestion
	if _, ok := doc.(*cdx.BOM); !ok {
		return errors.New("document is not a cyclonedx bom")
	}
	if err := encoder.EncodeVersion(doc.(*cdx.BOM), version); err != nil {
```

pkg/native/unserializers/unserializer_cdx.go
```diff
@@ -6,60 +6,102 @@ import (
...
+}
+
+func NewCDX(version, encoding string) *CDX {
+	return &CDX{
+		version:  version,
+		encoding: encoding,
+	}
+}
+
+// ParseStream reads a CycloneDX 1.5 from stream r usinbg the offcial CycloneDX
```
puerco, line 31: This comment needs to be updated to match the new function

pkg/native/serializer.go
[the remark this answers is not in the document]
puerco: I'm still not happy about unnecessarily breaking the api for personal preference but I'd rather move on than sit on this.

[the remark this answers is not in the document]
puerco: I'm still not happy about unnecessarily breaking the api for personal preference but I'd rather move on than sit on this.

puerco (approved):
OK, I think this is ready to go. I just found a few nits but really the big changes were reviewed in the PRs to the branch.

Great work on the options system @manifestori and, while I'm not very happy about some of the API changes and object restructures, I think we're ending up with a better codebase and I'd rather merg and resume development as we've been stuck on this one for too long. 

Thank you! :shipit:

puerco:
I'm going to go ahead and merge this PR. I'll address the nits in a small follow-up to enable us to cut a release.

PR #103, opened 2023-09-12
Support Lifecycles for CDX 1.5

manifestori:
This PR adds basic support for CDX 1.5, it reuses the CDX 1.5 unserializer and is built on top of the new serializers solution.

Closes most of the changes in #72

I do not like how the `unserializer` works as this is very nonergonomic, I strongly advise against "base" unserialize, but I did reuse cdx1.4 for simplicity. 

I think the discussion around "unserializers" and how can we make this reusable start on a new PR :)


api/sbom.proto
```diff
@@ -150,6 +151,22 @@ message Tool {
...
+
+    enum SBOMType {
+        OTHER = 0;
+        DESIGN = 1;
+        SOURCE = 2;
+        BUILD = 3;
+        ANALYZED = 4;
+        DEPLOAYED = 5;
+        RUNTIME = 6;
+    }
```
puerco, line 167: To keep the promise of lossless ingestion, my suggestion here is to also add the lifecycle phases that don't map to the community types.

Ingesting a cdx doc and rewriting it again should be lossless, so we need those types. Data loss (and the `OTHER` type) would only come into play when using the spdx sinks or older cdx.

Based on your mapping function and on the previous discussions on lifecycles vs sbom types, I think this should do (also watch out for the typo on `DEPLOYED`). Also note the mapping notes on the function below.

```suggestion
    enum SBOMType {
        OTHER = 0;
        DESIGN = 1; // design
        SOURCE = 2; // pre-build
        BUILD = 3; // build
        ANALYZED = 4; // post-build
        DEPLOYED = 5; // operations
        RUNTIME = 6;
        DISCOVERY = 7;  // CDX, no clear match
        DECOMISSION = 8;   // CDX, no clear match
    }
```

pkg/native/unserializers/unserializer_cdx15.go
```diff
@@ -0,0 +1,115 @@
...
+	case cdx.LifecyclePhaseDesign:
+		return sbom.DocumentType_DESIGN.Enum()
+	case cdx.LifecyclePhaseDiscovery:
+		return sbom.DocumentType_OTHER.Enum()
+	case cdx.LifecyclePhaseOperations:
+		return sbom.DocumentType_OTHER.Enum()
+	case cdx.LifecyclePhasePostBuild:
+		return sbom.DocumentType_OTHER.Enum()
+	case cdx.LifecyclePhasePreBuild:
+		return sbom.DocumentType_OTHER.Enum()
```
puerco, line 111:
```suggestion
		return sbom.DocumentType_SOURCE.Enum()
```

pkg/native/unserializers/unserializer_cdx15.go
```diff
@@ -0,0 +1,115 @@
...
+	case cdx.LifecyclePhaseDecommission:
+		return sbom.DocumentType_OTHER.Enum()
+	case cdx.LifecyclePhaseDesign:
+		return sbom.DocumentType_DESIGN.Enum()
+	case cdx.LifecyclePhaseDiscovery:
+		return sbom.DocumentType_OTHER.Enum()
+	case cdx.LifecyclePhaseOperations:
+		return sbom.DocumentType_OTHER.Enum()
+	case cdx.LifecyclePhasePostBuild:
+		return sbom.DocumentType_OTHER.Enum()
```
puerco, line 109:
```suggestion
	case cdx.LifecyclePhasePostBuild:
		return sbom.DocumentType_ ANALYZED.Enum()
```

pkg/native/unserializers/unserializer_cdx15.go
```diff
@@ -0,0 +1,115 @@
...
+	case cdx.LifecyclePhaseBuild:
+		return sbom.DocumentType_BUILD.Enum()
+	case cdx.LifecyclePhaseDecommission:
+		return sbom.DocumentType_OTHER.Enum()
+	case cdx.LifecyclePhaseDesign:
+		return sbom.DocumentType_DESIGN.Enum()
+	case cdx.LifecyclePhaseDiscovery:
+		return sbom.DocumentType_OTHER.Enum()
+	case cdx.LifecyclePhaseOperations:
+		return sbom.DocumentType_OTHER.Enum()
```
puerco, line 107:
```suggestion
	case cdx.LifecyclePhaseOperations:
		return sbom.DocumentType_DEPLOYED.Enum()
```

pkg/native/unserializers/unserializer_cdx15.go
```diff
@@ -0,0 +1,115 @@
...
+	decoder := cdx.NewBOMDecoder(r, cdx.BOMFileFormatJSON)
+	if err := decoder.Decode(bom); err != nil {
+		return nil, fmt.Errorf("decoding cyclonedx: %w", err)
+	}
+
+	md := &sbom.Metadata{
+		Id:      bom.SerialNumber,
+		Version: fmt.Sprintf("%d", bom.Version),
+		// Name:    ,
+		Date:          &timestamppb.Timestamp{},
```
puerco, line 32: nit: Capture the date here (strongly suggested) or leave it as nil (preferably not) otherwise it will revert to the unix epoch

puerco (commented):
Looks awesome @manifestori , just two suggestions about the mapping of the lifecycles and about adding the CDX types to the proto, otherwise looks good to me

api/sbom.proto
manifestori: Thank you for your help! I was having trouble converting to/from CDX according to the CISA document. After narrowing it down to a minimum of 6, I agree with your suggestion. we should include the CDX specific enums (discovery/decomission)

manifestori: @puerco so what do we need "OTHER" for? if we go "inclusive" (adding "DISCOVERY" and "DECOMISSION"). 
In my implementation, if a "phase" is OTHER, I still use the "phase name" (eg. "discovery") as a phase. 

Looking at https://github.com/spdx/spdx-3-model/blob/f73b70d8cd86ead09902937cbfdfdcd93af3ef1c/model/Core/Vocabularies/LifecycleScopeType.md#L4

SPDX 3 lifecycles are these:
```
design: TODOdescription
build: TODOdescription
development: TODOdescription
test: TODOdescription
runtime: TODOdescription
other: TODOdescription
```

I assume we should "save" the OTHER prop for "non-phase" in CDX, so we can go from/to CDX1.5/SPDX3.
However, there is a big difference between usage for "lifecycle" CDX in metadata, and SPDX as a relationship type. We will be able to translate those, but I'm pretty sure we would lose the "description" when going from CDX to SPDX.

CDX 1.5
```json
 "lifecycles": [
      {
        "phase": "build"
      },
      {
        "name": "platform-integration-testing",
        "description": "Integration testing specific to the runtime platform"
      }
    ],
```

I suggest we discuss this briefly in today's Cohort meeting

pkg/native/unserializers/unserializer_cdx15.go
manifestori: Thanks!

manifestori: Should I do it for CDX 1.4 as well?

api/sbom.proto
puerco: Yeah in my view, `OTHER` should be the default when unknown data makes it into the field (which should never happen via the unserializers). It should be the value populated when using the struct option in CDX. In other enums we use `UNKNOWN` but here I think OTHER is fine as it can also point to.

Regarding the SPDX3 values, you are confusing the lifecycles of the build profile with the SBOM Type. This is the one you should be looking at:

https://github.com/spdx/spdx-3-model/blob/f73b70d8cd86ead09902937cbfdfdcd93af3ef1c/model/Software/Vocabularies/SbomType.md

All the build profile stuff is out of the scope of protobom, so don't worry about it :)

pkg/native/unserializers/unserializer_cdx15.go
puerco: Cool, but we can do it in a follow up too.

pkg/native/serializers/serializer_cdx.go
```diff
@@ -68,6 +69,21 @@ func (s *SerializerCDX) Serialize(bom *sbom.Document, _ *native.SerializeOptions
...
+	for _, dt := range bom.Metadata.DocumentTypes {
+		var lfc cdx.Lifecycle
+
+		if dt.Type == nil {
+			lfc.Name = *dt.Name
+			lfc.Description = *dt.Description
+		} else if *dt.Type == sbom.DocumentType_OTHER {
+			lfc.Phase = cdx.LifecyclePhase(strings.ToLower(*dt.Name))
+		} else if *dt.Type != sbom.DocumentType_OTHER {
+			lfc.Phase = cdx.LifecyclePhase(strings.ToLower(dt.Type.String()))
```
manifestori, line 81: @puerco that's where I handle the different cases: mapped phase, unmapped phase (OTHER), and custom object.

api/sbom.proto
puerco: No no, don't do that. If the lifecycle is in the enum and you move it to the opt2 then you are effectively modifying the document before you are even finished reading it.

When ingesting data into enums, we grow the options to handle the vocabularies of the formats to not lose any data. See the hash algorithms, for example: There are common algorithms and then there are others that are specific to SPDX. Regardless they all exist in the enum.

manifestori: Oh ok, Ill do that then!

PR #104, opened 2023-09-13
Configurable serializers

manifestori:
This PR proposes a new breaking change that will allow users to both set config on instantiation and define them on the fly. 

This allows specific serializer options to be defined and implemented by the concrete serializer, allowing you to pass them through the writer. 

The alternative is to implement this with options pattern passing (...SerializerOption) instead of interface{}
The serializer would define its specific (or shared) "WithXXX" functions and they could passed through the writer (again, without tempering with them)

### DIFF
--- a/pkg/writer/writer.go
+++ b/pkg/writer/writer.go
@@ -13,23 +13,28 @@ import (
 )
 
 type Writer struct {
-	Config *Config
+	RenderOptions   map[string]*native.RenderOptions
+	SerialzeOptions map[string]*native.SerializeOptions
+	Format          formats.Format
 }
 
 var (
-	regMtx        sync.RWMutex
-	serializers   = make(map[formats.Format]native.Serializer)
-	defaultConfig = &Config{
-		RenderOptions: &DefaultRenderOptions{
+	regMtx               sync.RWMutex
+	serializers          = make(map[formats.Format]native.Serializer)
+	defaultRenderOptions = &native.RenderOptions{
+		CommonRenderOptions: native.CommonRenderOptions{
 			Indent: 4,
 		},
-		SerializeOptions: &DefaultSerializeOptions{},
 	}
+	defaultSerializeOptions = &native.SerializeOptions{}
+	defaultFormat           = formats.CDX15JSON
 )
 
 func New(opts ...WriterOption) *Writer {
 	r := &Writer{
-		Config: defaultConfig,
+		RenderOptions:   make(map[string]*native.RenderOptions),
+		SerialzeOptions: make(map[string]*native.SerializeOptions),
+		Format:          defaultFormat,
 	}
 
 	for _, opt := range opts {
@@ -74,9 +79,14 @@ func GetFormatSerializer(format formats.Format) (native.Serializer, error) {
 	return nil, fmt.Errorf("no serializer registered for %s", format)
 }
 
-func (w *Writer) WriteStreamWithConfig(bom *sbom.Document, format formats.Format, config *Config, wr io.WriteCloser) error {
+func (w *Writer) WriteStreamWithOptions(bom *sbom.Document, wr io.WriteCloser, o *Options) error {
 	if bom == nil {
 		return fmt.Errorf("unable to write sbom to stream, SBOM is nil")
+	}
+
+	format := o.Format
+	if o.Format == "" {
+		format = w.Format
 	}
 
 	serializer, err := GetFormatSerializer(format)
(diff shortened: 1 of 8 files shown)


veramine:
I test protobom right now using a small tool that is roughly the following:

```
	document := sbom.NewDocument()
	...
	// configure metadata
	// configure nodes
	// add nodes to NodeList
	// set root node
	...
	file, err := os.CreateTemp("", "*.cdx.json")
	..
	w := writer.New()
	err = w.WriteStream(document, file)
	..
```

Super duper simple.  

Can we make WriteStream() still work this way without any configuration if I am OK with default settings?  Is there a default Config that will be used if I don't specify one?

pkg/writer/writer.go
```diff
@@ -72,36 +74,50 @@ func GetFormatSerializer(format formats.Format) (native.Serializer, error) {
...
+		ro = w.Config.RenderOptions
+	}
+	if err := serializer.Render(nativeDoc, wr, ro); err != nil {
 		return fmt.Errorf("writing rendered document to string: %w", err)
 	}
 
 	return nil
 }
 
+func (w *Writer) WriteStream(bom *sbom.Document, format formats.Format, wr io.WriteCloser) error {
```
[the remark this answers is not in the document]
veramine: aha, yes that was what I was wondering in my question.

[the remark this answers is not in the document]
veramine: Maybe you could take that approach and then modify writer_test.go to use the backwards compatible WriteStream()

veramine (changes requested):
WriteStream() backwards compatibility/default config, per comments above

pkg/writer/writer.go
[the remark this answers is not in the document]
manifestori: Yes, I'll make sure to not break the WriteStream/File functions only adding new functions to support passing formats/serilizerconfig

manifestori:
Already in progress, I'll push my changes tomorrow, I've been discussed @puerco about good ways to do this. I think we're nearly there :)

manifestori:
@veramine this is now ready for review, I've nailed down backward compatibility for WriteStream/WriteFile.

@houdini91 Serailizer options: both for Render and Serialize (which are two configurable options for a Serializer) now have ways to be set on demand, by default, or by instantiating them.

@puerco according to your suggestion - we now have a "map" of options per serializer. this map can be configurable, further helpers and better instantiation can be improved in future work.

Also, options now have "Common" options which are preconfigured options for serialize/render, such as "Indent", a user may also pass arbitrary options via the "Option" prop which is an "interface{}", in the serializer a user would be able to cast into a specific type (in case of "custom" serializer) and use non-official options.

I hope this qualifies as it covers all the cases, tests are 96%+ coverage (only small error checks we're not testing, which shouldn't happen).

if someone thinks this options solution is no-go please, lets discuss this promptly 🥂

pkg/writer/writer.go
```diff
@@ -72,36 +79,90 @@ func GetFormatSerializer(format formats.Format) (native.Serializer, error) {
...
 	}
 
-	if err := serializer.Render(nativeDoc, wr, &native.RenderOptions{
-		Indent: w.Indent,
-	}); err != nil {
+	ro := o.RenderOptions
+	if ro == nil {
+		ro = w.RenderOptions[key]
+	} else {
+		fmt.Println("its empty!")
```
puerco, line 112: Should we remove this print?
```suggestion
```

pkg/writer/writer.go
```diff
@@ -13,21 +13,28 @@ import (
...
-	defaultFormat = formats.CDX15JSON
+	regMtx               sync.RWMutex
+	serializers          = make(map[formats.Format]native.Serializer)
+	defaultRenderOptions = &native.RenderOptions{
+		CommonRenderOptions: native.CommonRenderOptions{
+			Indent: 4,
+		},
+	}
+	defaultSerializeOptions = &native.SerializeOptions{}
+	defaultFormat           = formats.CDX15JSON
```
puerco, line 30: I don't have a strong opinion here but perhaps we should seek consensus on moving the default to CDX 1.5. At least check with the cohort to see if there are any objections.

puerco (approved):
This looks great, @manifestori . Thank you! It captures the discussions we've been having very well and I also appreciate many other ideas you added improving even more the plan we were discussing.

I will do a full pass later today but after looking at it, I don't see anything blocking it, just a few nits here and there which I think can be handled after merging. Since this is merging into #102 I think we should go ahead with it and handle any nits and other potential comments over there.

Awesome work :rocket:

pkg/writer/writer.go
veramine: Probably best to use our "most complete" serializer as the default choice.  That's unlikely to be CDX 1.5 at the moment.  I've been working on SPDX 2.3 mostly and that's looking pretty good.  I haven't looked at the CDX 1.5 support at all yet to see how complete our support is.

pkg/writer/writer.go
```diff
@@ -72,36 +79,90 @@ func GetFormatSerializer(format formats.Format) (native.Serializer, error) {
...
+		fmt.Println("its empty!")
+	}
+	if err := serializer.Render(nativeDoc, wr, ro); err != nil {
 		return fmt.Errorf("writing rendered document to string: %w", err)
 	}
 
 	return nil
 }
 
+func (w *Writer) WriteStream(bom *sbom.Document, wr io.WriteCloser) error {
```
veramine, line 121: Yesssss lets go!  Thank you!

veramine (approved):
This looks great!  Thanks!  I appreciate you adding in the WriteStream() function back in for the laaaaazy devs :) :) :)

houdini91:
Lgtm

PR #106, opened 2023-09-15
Add NodeSiblings and NodeGraph to NodeList

puerco:
This PR adds two new functions to query data:

* `nodelist.NodeSiblings()` takes a node id and returns a new NodeList with it at the root and all its first-level siblings with its edges.
* `nodelist.NodeGraph()` returns a new nodelist with node `id` at the top, extracting and preserving its graph fragment.

This PR does affect the serializer/unserializing API, only the external querying functions. Full tests are included.

### DIFF
--- a/pkg/sbom/nodelist.go
+++ b/pkg/sbom/nodelist.go
@@ -629,3 +629,123 @@ func (nl *NodeList) reconnectOrphanNodes() {
 		}
 	}
 }
+
+// NodeGraph looks for node id and returns a new NodeList with its full dependency
+// graph. NodeGraph will traverse the SBOM graph and add all nodes connected to id.
+func (nl *NodeList) NodeGraph(id string) *NodeList {
+	nodelist := &NodeList{
+		Nodes:        []*Node{},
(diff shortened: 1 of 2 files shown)


veramine (approved):
This looks fine to me.  I tested the change with a small SBOM having a flat nodelist with a few hundred peer nodes.  I tested both NodeGraph(id) and NodeSiblings(id) and both functions returned me a functional NodeList with just the one node I passed in as the root.  If that's what you wanted, it worked for me!  I was kinda thinking I would get all the nodes in my original nodelist..  but I don't actually know the intended use case for this function so if it is working as you intended it to, totally fine with me.

puerco:
I think you are describing the intended behavior. In this example, if you query for B you only get B:

```mermaid
flowchart LR
    root-->B
   root-->C
```

To also get the rootElement C in the same query it then you would need an edge between B and C as traversing stops at the nodelist top:

```mermaid
flowchart LR
    root-->B
   root-->C
  B-->C
```

veramine:
okay yep!  LGTM!

PR #108, opened 2023-09-19
Ignore CreationInfo if not present

veramine:
Fixes an issue found by fuzzing (https://github.com/bom-squad/sbom-convert/pull/13).  Added check for nil CreationInfo before trying to use it. 

```
if spdxDoc.CreationInfo != nil {
...
}
```


jspeed-meyers (approved):
LGTM. Thanks for fixing this bug. 🐛

PR #109, opened 2023-09-19
Return error if native serializer does not return valid root node

veramine:
Fixes an issue found by fuzzing (https://github.com/bom-squad/sbom-convert/pull/13). Added check to return error if serializer root component is nil.

```
	rootComponent, err := s.root(ctx, bom)
	if err != nil {
		return nil, fmt.Errorf("generating SBOM root component: %w", err)
	} else if rootComponent == nil {
		return nil, fmt.Errorf("No SBOM root component")
	}
```

(the else if is added)


jspeed-meyers (approved):
Nice. I love this feedback loop from @wei-deepbits's fuzzer findings back to `protobom`. Great work all!

Issue #112, opened 2023-10-04
pkg/formats/sniffer.go can't detect CycloneDX files if they are minified to one line


sweoggy:
As title, if I have a file like this

`{"bomFormat":"CycloneDX","specVersion":"1.4","version":1,"serialNumber":"urn:uuid:b35ea91e-91c2-40ae-97fe-c015a4fd8790","metadata":{...`

sniff() will not detect the format and SniffReader() will return "unknown SBOM format".

If I instead prettify the document so it start likes this
```json
{
    "bomFormat": "CycloneDX",
    "specVersion": "1.4",
    "version": 1,
    "serialNumber": "urn:uuid:b35ea91e-91c2-40ae-97fe-c015a4fd8790",
    "moreKeys": {},
}
```

it works just fine. It would be good if minified JSON files worked too :)

Thanks for an excellent library!

veramine:
Thanks @sweoggy !  We were using some goofy string hacks to get the version previously.  I just submitted a PR (https://github.com/bom-squad/protobom/pull/113) to use the json Unmarshaler instead for JSON-based SBOMs.  I also added your minified.cdx.json to the test cases we will ensure work going forward.  Will let you know when this PR gets reviewed and merged.

PR #113, opened 2023-10-04
Use json.Unmarshal to get CDX specVersion (replacing string hacks)

veramine:
Attempt to read just the SpecVersion from the JSON via json.Unmarshal.  If this succeeds, use that version.  If it fails, use the string hacks in place already.

Added a test case for the minified json included in the issue.

Fixes https://github.com/bom-squad/protobom/issues/112

### DIFF
--- a/pkg/formats/sniffer.go
+++ b/pkg/formats/sniffer.go
@@ -2,6 +2,7 @@ package formats
 
 import (
 	"bufio"
+	"encoding/json"
 	"fmt"
 	"io"
 	"os"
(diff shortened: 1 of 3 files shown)


jspeed-meyers (approved):
Thanks for adding this test and fixing it up. LGTM!

puerco (approved):
Thanks @veramine, I think we can expand the same struct to also capture the SPDX version string in one go. lets merge this one and we can expand it in a follow-up

PR #114, opened 2023-10-05
Capture the SPDX version and CDX version both in one shot

veramine:
I'm proposing to revert the change I made yesterday to fix https://github.com/bom-squad/protobom/issues/112.  This implementation is better as it will capture both the SPDX and CDX versions in one shot without having to do any string hacks. Thanks for suggestion @puerco

### DIFF
--- a/pkg/formats/sniffer.go
+++ b/pkg/formats/sniffer.go
@@ -138,7 +138,6 @@ func (st *sniffState) Format() Format {
 type cdxSniff struct{}
 
 func (c cdxSniff) sniff(data []byte) Format {
-
 	// protobom only supports CDX formats as JSON
 	//  we are parsing the JSON in SniffReader by decoding to the SpecVersionStruct
 	//   removing all the previous JSON-related string matching from this function
--- a/pkg/formats/sniffer.go
+++ b/pkg/formats/sniffer.go
@@ -86,7 +86,7 @@ func (fs *Sniffer) SniffReader(f io.ReadSeeker) (Format, error) {
 
 	_, err = f.Seek(0, 0)
 	if err != nil {
-		return fmt.Errorf("seeking to the beginning of SBOM file: %w", err)
+		return "", fmt.Errorf("seeking to the beginning of SBOM file: %w", err)
 	}
 
 	fileScanner := bufio.NewScanner(f)
--- a/pkg/formats/sniffer.go
+++ b/pkg/formats/sniffer.go
@@ -86,7 +86,7 @@ func (fs *Sniffer) SniffReader(f io.ReadSeeker) (Format, error) {
 
 	_, err = f.Seek(0, 0)
 	if err != nil {
-		fmt.Printf("WARNING: could not seek to beginning of file: %v", err)
+		return fmt.Errorf("seeking to the beginning of SBOM file: %w", err)
 	}
 
 	fileScanner := bufio.NewScanner(f)
--- a/pkg/formats/sniffer.go
+++ b/pkg/formats/sniffer.go
@@ -138,27 +138,14 @@ func (st *sniffState) Format() Format {
 type cdxSniff struct{}
 
 func (c cdxSniff) sniff(data []byte) Format {
-	state := getSniffState(CDXFORMAT)
 
-	stringValue := string(data)
-	if strings.Contains(stringValue, `"bomFormat"`) && strings.Contains(stringValue, `"CycloneDX"`) {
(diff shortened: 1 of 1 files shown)


pkg/formats/sniffer.go
```diff
@@ -46,6 +46,49 @@ func (fs *Sniffer) SniffReader(f io.ReadSeeker) (Format, error) {
...
+		if specversionjson.BomFormat == "CycloneDX" {
+			if specversionjson.CDXSpecVersion == "1.3" {
+				return CDX13JSON, nil
+			} else if specversionjson.CDXSpecVersion == "1.4" {
+				return CDX14JSON, nil
+			} else if specversionjson.CDXSpecVersion == "1.5" {
+				return CDX15JSON, nil
+			} else {
+				// JSON + BomFormat CycloneDX but specVersion not 1.3, 1.4, or 1.5
+				return "", fmt.Errorf("unknown SBOM format")
```
puerco, line 70: If #102 merges before this one, you will be able to replace this with:
```suggestion
   f, err := cdxformats.ParseVersion(specversionjson.CDXSpecVersion)
   if err != nil {
     return "", fmt.Erorf("parsing cyclonedx format: %w", err)
   }
   return f, nil
```

(totally fine to this later)

pkg/formats/sniffer.go
```diff
@@ -46,6 +46,49 @@ func (fs *Sniffer) SniffReader(f io.ReadSeeker) (Format, error) {
...
+		CDXSpecVersion  string `json:"specVersion"`
+		SPDXSpecVersion string `json:"spdxVersion"`
+	}
+
+	decoder := json.NewDecoder(f)
+
+	var specversionjson SpecVersionStruct
+	err := decoder.Decode(&specversionjson)
+	if err == nil {
+		if specversionjson.BomFormat == "CycloneDX" {
```
puerco, line 61:
```suggestion
		if strings.ToLower(specversionjson.BomFormat) == CDXFORMAT {
```

pkg/formats/sniffer.go
```diff
@@ -46,6 +46,49 @@ func (fs *Sniffer) SniffReader(f io.ReadSeeker) (Format, error) {
...
+				return "", fmt.Errorf("unknown SBOM format")
+			}
+		}
+	}
+
+	// not JSON.  Parse line-by-line with string hacks
+
+	_, err = f.Seek(0, 0)
+	if err != nil {
+		fmt.Printf("WARNING: could not seek to beginning of file: %v", err)
```
puerco, line 89: Dont print to STDOUT in the library. In fact, I think we should error here as a failure to rewind the pointer is going to end up in an error parsing the SBOM anyway,
```suggestion
		return fmt.Errorf("seeking to the beginning of SBOM file: %w", err)
```

pkg/formats/sniffer.go
```diff
@@ -103,27 +146,13 @@ func (c cdxSniff) sniff(data []byte) Format {
...
-					state.Version = ver
-					state.Encoding = JSON
-				}
+	if strings.Contains(stringValue, `"specVersion"`) {
+		parts := strings.Split(stringValue, ":")
+		if len(parts) == 2 {
+			ver := strings.TrimPrefix(strings.TrimSuffix(strings.TrimSuffix(strings.TrimSpace(parts[1]), ","), "\""), "\"")
+			if ver != "" {
+				state.Version = ver
+				state.Encoding = JSON
```
puerco, line 155: Now that we've handled JSON parsing using the universal struct, I think we could remove all the code that handles JSON in the string parser.

We can do this in a follow up too.

puerco (commented):
Thanks @veramine, a few comments 👇

pkg/formats/sniffer.go
veramine: yes, thank you.  Fixed.

pkg/formats/sniffer.go
veramine: Yes, good idea.  I added a commit to remove the strings-based parsing from both sniff functions - the entire content of the cdxSniff version and the portion of spdxSniff.  Thanks!

pkg/formats/sniffer.go
veramine: yeah, good idea

pkg/formats/sniffer.go
veramine: Sure, as soon as #102 merges I'll make that change, thanks!

puerco (approved):
Thanks @veramine!

PR #119, opened 2023-10-11
Streamline and colorize README examples

puerco:
This PR streamlines the main README examples and colorizes them for easy reading. It also adds comments to better understand what they are doing. It also adds the missing relationships in the second example.

Note that the second example hits a bug in the new writer implementation, I'm fixing it now.

/cc @veramine @jspeed-meyers


veramine:
Beautiful!  Thank you!

Issue #62, opened 2023-08-08
Support CycloneDX 1.5


manifestori:
As the spec for CycloneDX 1.5 was released, and some OSS generators already use it by default (Trivy, cdxgen)

Also, `cyclonedx-go` lib is being upgraded to support 1.5 here: https://github.com/CycloneDX/cyclonedx-go/tree/spec-v1.5

I want to support converting from and to cdx 1.5

TODO: 

- [x] Implement unserializer for cdx1.5
- [x] Implement serializer for cdx1.5
- [x] Update supporting lib, refactor common functionalities between cdx formats
- [ ] Update tests, docs, and examples
- [ ] Open PR for `sbom-convert` and update the supported format list

manifestori:
I've researched `cyclonedx-go` and realized how they are doing backward/forward compatibility. 

This is true for 1.x and might change when/if they break this notion with 2.x. 

Decoding: 
Marshaling uses go annotations to convert from xml/json to go structs. 
The spec itself is backward compatible, so if, for example, a CDX1.3 is read, it won't have the "vulnerabilities" prop allocated.
: 
Encoding: 
When we encode back to `cyclonedx-go` does a "copy and convert" which first copy into a new bom (to preserve immutability) and then, fix compatibility issues on the fly, if for example:

```
if b.Metadata != nil {
		if specVersion < SpecVersion1_3 {
			b.Metadata.Licenses = nil
			b.Metadata.Properties = nil
		}
..
```

I have been uncertain about how they will implement backward/forward compatibility for changes in the schema that resolves with a type change. The only instance this happens is with Tools (#68) 
Note the schema changes: https://github.com/CycloneDX/specification/pull/198

Their option is to override UnmarshalJSON and fix older spec SBOM from a slice to an object, but this is still an open question: Tracking it [[here](https://github.com/CycloneDX/cyclonedx-go/issues/115)](https://github.com/CycloneDX/cyclonedx-go/issues/115)

manifestori:
> As the spec for CycloneDX 1.5 was released, and some OSS generators already use it by default (Trivy, cdxgen)
> 
> Also, `cyclonedx-go` lib is being upgraded to support 1.5 here: https://github.com/CycloneDX/cyclonedx-go/tree/spec-v1.5
> 
> I want to support converting from and to cdx 1.5
> 
> TODO:
> 
> * [ ]  Implement unserializer for cdx1.5
> * [ ]  Implement serializer for cdx1.5
[...]

PR #121, opened 2023-10-11
Update Supported Formats and Version Table

jspeed-meyers:
CycloneDX 1.5 is supported now, right?

### DIFF
--- a/README.md
+++ b/README.md
@@ -26,7 +26,7 @@ the golang library.
 | SPDX | 2.3 | tag-value | planned | - |
 | SPDX | 3.0 | JSON | planned | planned |
 | CycloneDX | 1.4 | JSON | supported | supported |
-| CycloneDX | 1.5 | JSON | planned | planned |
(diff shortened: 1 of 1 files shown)


veramine:
Yes, I'm absolutely able to generate a CDX1.5 SBOM via writer.New(writer.WithFormat(formats.CDX15JSON)):

```
{
  "$schema": "http://cyclonedx.org/schema/bom-1.5.schema.json",
  "bomFormat": "CycloneDX",
  "specVersion": "1.5",
...
```

But Ori was tracking some issues with the official go library.  Or maybe he was being asked to make changes in the official go library?  I don't remember exactly.  Please check with him and if he says yes than this is a yes for sure!

veramine:
Here was Ori's issue https://github.com/bom-squad/protobom/issues/62

jspeed-meyers:
cc @manifestori -- It looks like 1.5 landed, right?

puerco (approved):
Yes it has landed, we can still improve it some more but it is in! :shipit:

PR #123, opened 2023-10-11
Reorganize options data structures

### LINKED ISSUE #122: Inconsistent behavior with new options structs
After testing for the past two days the new option structs in our applications that use protobom I've notivecd that we introduced a very cumbersome usability problem in the way the options are structured.

Common options are not really common and the writer's functions are not really the writer's. This means that for every format we have a separate set of `CommonRenderOptions` and `CommonSerializeOptions`. This is not practical. Take this snippet for example:

```golang

    w := writer.New()

    // Write the SBOM to STDOUT in SPDX 2.3:
    w.Format = formats.SPDX23JSON
    w.WriteStream(document, os.Stdout)

    // Write the SBOM to STDOUT in CycloneDX 1.4:
    w.Format = formats.CDX14JSON
    w.WriteStream(document, os.Stdout)
```

In this example, each call to `w.WriteStream` has a different options set, this means that we will need to reconfigure with the same options everything from indetntation, logging, deprecation log, etc whenever we change formats.  If not, renders for each format will behave differently

This is because the option sets for each format are keyed at the wrong place, for each format we repeat the full writer option set:

https://github.com/bom-squad/protobom/blob/b3dae45f586c6bfd9da72786a563b9fc7476a1c5/pkg/writer/writer.go#L35-L36

This also introduces all sorts of usability problems in practice. For example. Before, if I wanted to set the indentation level of the output, I could just do:

```golang
// Original way of setting options
w := writer.New()
w.Options.Indent = 4
```
The new equivalent becomes this to avoid a panic:

```golang
w := writer.New()
if _, ok := w.RenderOptions[string(w.Format)]; ok {
   w.RenderOptions[string(w.Format)].Indent = 4
}
```

But there is a problem, that code introduces inconsistency across all sets of options. So, to force a consistent behavior I need to change each individual options set:

```golang
w := writer.New()
for f = range w.RenderOptions {
   w.RenderOptions[f].Indent = 4
}
```
This is clearly inefficient and impractical. We need to factor out common options into a reusable struct to be able to set the common options in a practical way (as in the first example).

The only sets that we need to catalog behind a key are the serializer functions, so I propose we restructure the functions like this:

```golang
// Current writer options, two sets repeated for each format

r := &Writer{
    RenderOptions: : map[string]*native.RenderOptions{
        string(formats.SPDX23JSON): native.RenderOptions{
                CommonRenderOptions {}
                // more optios
        },
        string(formats.CDX14JSON): native.RenderOptions{
                CommonRenderOptions {} // these common options repeat the struct above
                // more optios
        },
    },   
    SerializeOptions: map[string]*native.SerializeOptions{
        string(formats.SPDX23JSON): native.SerializeOptions{
                CommonSerializeOptions
                // more options
        },
        string(formats.CDX14JSON): native.SerializeOptions{
                CommonSerializeOptions // same options as the spdx23 set
                // more options
        },
    },
}

// New Proposal, one options struct with the writer options, embedding the 
// common serializer and render options and a dict keyed to the specific 
// serializer structs:

type Options struct {
	Format           formats.Format
	RenderOptions    *native.RenderOptions
	SerializeOptions *native.SerializeOptions
	formatOptions    map[string]interface{
                 string(formats.SPDX23JSON): spdxOpts{},
                 string(formats.CDX14JSON): spdxOpts{},
        }
}

```

This new options struct lets us control a unique shared setting for the serializer options, no matter if it is in the render or serialize structs and it is effective for al calls to serialize documents:

```
  myWriter.Options.Indent = 4 
  myWriter.Options.Logger = myLogger
  myWriter.Options.DebugLevel = 1
  myWriter.Options.ErrOnDataLoss = true
```
I will send a PR to demo the proposal, we can discuss it in our meeting.

puerco:
This PR restructures the options data structures to address the problems delineated in #122. It is a set of changes to improve the usability and consistency of the writer object and its option. The changes are as follows:

* It modifies the writer to have a single options struct that embeds a single instance of the render and serialize options while keeping a catalog to host each of the serializer driver functions.
* The serializer functions now take a set of common options (render and serialize options) and the native serializer options.
* The fakes have been updated accordingly.
* The writer options now have functions to set and get each driver's options set.
* Tests are updated to handle the new universal options struct

The new options struct looks as follows:

```golang
type Options struct {
	// Writer Options:
	Format           formats.Format
	// FutureOption1
	// FutureOption2

	// Embedded Render options
	RenderOptions    *native.RenderOptions 

	// Embedded  Serialize Options
	SerializeOptions *native.SerializeOptions

	// Serializer Dirver native options catalog
	formatOptions    map[string]interface{
                 string(formats.SPDX23JSON): spdxOpts{},
                 string(formats.CDX14JSON): spdxOpts{},
        }
}
```

Fixes #122

### DIFF
--- a/pkg/writer/writer.go
+++ b/pkg/writer/writer.go
@@ -104,7 +104,7 @@ func (w *Writer) WriteStreamWithOptions(bom *sbom.Document, wr io.WriteCloser, o
 		so = defaultOptions.SerializeOptions
 	}
 
-	nativeDoc, err := serializer.Serialize(bom, so)
+	nativeDoc, err := serializer.Serialize(bom, so, o.GetFormatOptions(serializer))
 	if err != nil {
 		return fmt.Errorf("serializing SBOM to native format: %w", err)
 	}
@@ -114,7 +114,7 @@ func (w *Writer) WriteStreamWithOptions(bom *sbom.Document, wr io.WriteCloser, o
 		ro = defaultOptions.RenderOptions
 	}
 
-	if err := serializer.Render(nativeDoc, wr, ro); err != nil {
+	if err := serializer.Render(nativeDoc, wr, ro, o.GetFormatOptions(serializer)); err != nil {
 		return fmt.Errorf("writing rendered document to string: %w", err)
 	}
 
--- a/pkg/writer/writer.go
+++ b/pkg/writer/writer.go
@@ -14,33 +14,31 @@ import (
 )
 
 type Writer struct {
-	RenderOptions    map[string]*native.RenderOptions
-	SerializeOptions map[string]*native.SerializeOptions
-	Format           formats.Format
+	Options *Options
 }
 
 var (
-	regMtx               sync.RWMutex
-	serializers          = make(map[formats.Format]native.Serializer)
-	defaultRenderOptions = &native.RenderOptions{
-		CommonRenderOptions: native.CommonRenderOptions{
+	regMtx         sync.RWMutex
+	serializers    = make(map[formats.Format]native.Serializer)
+	defaultOptions = &Options{
+		RenderOptions: &native.RenderOptions{
 			Indent: 4,
 		},
+		SerializeOptions: &native.SerializeOptions{},
+		formatOptions:    map[string]interface{}{},
 	}
-	defaultSerializeOptions = &native.SerializeOptions{}
 )
 
 func New(opts ...WriterOption) *Writer {
-	r := &Writer{
-		RenderOptions:    make(map[string]*native.RenderOptions),
-		SerializeOptions: make(map[string]*native.SerializeOptions),
+	w := &Writer{
+		Options: defaultOptions,
 	}
 
 	for _, opt := range opts {
-		opt(r)
+		opt(w)
 	}
 
-	return r
+	return w
 }
 
 func init() {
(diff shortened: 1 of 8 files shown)


veramine:
This change makes sense to me, yes.  I just tested it and didn't need to change any of my test tools, everything still just worked the same as before.  Thank you for that!  Where are the defaults set?

veramine:
Okay,  I see the one default for RenderOptions Indent:4.  Are there any other defaults?  

This change seems fine to me.  The only thing I was thinking is dang are we really doing all this Options infrastructure for RenderOptions and SerializeOptions and 23487263487632 FormatOptions when really we are just trying to give the user control over indentation?  Did you have in mind a whole bunch of other potential options in the future?  If so, let me add those to the README for people who might have the same idea at first glance.  I want to add a 4th example to show Options usage.

I'm totally fine with the change though.  I'll approve.

veramine (approved):

puerco:
Ah no the indent setting is just a single example, we're about to start adding all the real options. These will control everything from indent levels to how we render the multi artifact cdx sboms or if you want SPDX3 in graph or embedded, etc.

pkg/writer/writer.go
```diff
@@ -14,33 +14,31 @@ import (
...
 	}
 
 	for _, opt := range opts {
-		opt(r)
+		opt(w)
 	}
 
-	return r
+	return w
 }
```
manifestori, line 42: Thanks <3

manifestori (approved):
LGTM. 

Looks great, I think thats a good long term solution 🚀

PR #125, opened 2023-10-12
Restructure reader options

puerco:
This PR restructures the reader options just as we did for the writer options in #123. It basically creates a single Options set that embeds the unserializer and a collection for the serializer driver options. It is a big PR but it is mostly harmless as we don't have any options defined yet and the design principles are the same as in #123:

```golang
type Options struct {
	Format             formats.Format
	UnserializeOptions: *native.UnserializeOptions {}
	formatOptions      map[string]interface{}
}
```

Note: This is a blocker for cutting v0.2.0 as it evens out the patterns in the readers and writers.

### DIFF
--- a/pkg/native/nativefakes/fake_unserializer.go
+++ b/pkg/native/nativefakes/fake_unserializer.go
@@ -10,11 +10,12 @@ import (
 )
 
 type FakeUnserializer struct {
-	UnserializeStub        func(io.Reader, *native.UnserializeOptions) (*sbom.Document, error)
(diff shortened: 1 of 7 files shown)


jspeed-meyers (approved):

veramine (approved):
This looks fine to me.  Smoke test with my test tools also worked fine.   It will be fun to see options being added based on this..

puerco:
Thanks for your reviews!

puerco:
Thanks for your reviews!

Issue #126, opened 2023-10-13
Syft-generated SPDX 2.3 SBOM cannot be written to CDX due to root node missing from node list


veramine:
Filing this on behalf of @wei-deepbits 

Protobom ingests Syft-generated SPDX 2.3 SBOMs just fine.  Here's an example:

[in.json](https://github.com/bom-squad/protobom/files/12888846/in.json)

In this case, the root node is SPDXID SPDXRef-DOCUMENT but it is not in the node list.  

```
{
 "spdxVersion": "SPDX-2.3",
 "dataLicense": "CC0-1.0",
 "SPDXID": "SPDXRef-DOCUMENT",
 "name": "/home/wei/code/repos/python/abhiTronix/vidgear",
 "documentNamespace": "https://anchore.com/syft/dir/home/wei/code/repos/python/abhiTronix/vidgear-710e3edc-b86e-4c5a-b71a-2f81c876b318",
 "creationInfo": {
  "licenseListVersion": "3.21",
  "creators": [
   "Organization: Anchore, Inc",
   "Tool: syft-0.85.0"
  ],
  "created": "2023-07-19T15:20:12Z"
 }
```

There are only two nodes in the node list:  SPDXRef-Package-python-pyzmq-d345c09cbb0c1e23 and SPDXRef-File-setup.py-c829033125510d5a.  But we need DOCUMENT to be in the list as well.  We try to find it in serializer_cdx.go here:

```golang
	if bom.NodeList.RootElements != nil && len(bom.NodeList.RootElements) > 0 {
		for _, id := range bom.NodeList.RootElements {
			// Search for the node and add it
			for _, n := range bom.NodeList.Nodes {
         ...
```

When we don't find it, the tree doesn't have a root and we end up here:

```golang
        if rootComponent == nil {
		return nil, fmt.Errorf("No SBOM root component")
	}
```

veramine:
I'm gonna try to figure this one out

veramine:
Fixed by https://github.com/bom-squad/protobom/pull/128

PR #128, opened 2023-10-13
Allow serialization of SBOM with no rootComponent

veramine:
With CycloneDX, the root node is expected to be in the node list.  However, in SPDX the Document section can be the root node with no package or file otherwise acting as the root node.  I had previously added a check in the CycloneDX serializer to reject protobom documents where the root node was not in the node list.  (https://github.com/bom-squad/protobom/pull/109)  That change does not allow some protobom documents originating from a valid SPDX document to be written to CycloneDX.  See https://github.com/bom-squad/protobom/issues/126.

Example originating SBOM:  
[in.json](https://github.com/bom-squad/protobom/files/12899027/in.json)

With this change, we allow CDX serialization of protobom Documents where the root node is not in the Node List.  I contemplated instead adding a fake node to the node list but decided not to.  Happy to reconsider that and explicitly add a node to the node list when it is missing and then reference that as the root if the community thinks that is the right approach.  

This change unblocks https://github.com/bom-squad/sbom-convert/pull/13

### DIFF
--- a/pkg/native/serializers/serializer_cdx.go
+++ b/pkg/native/serializers/serializer_cdx.go
@@ -62,11 +62,10 @@ func (s *CDX) Serialize(bom *sbom.Document, _ *native.SerializeOptions, _ interf
 	rootComponent, err := s.root(ctx, bom)
 	if err != nil {
 		return nil, fmt.Errorf("generating SBOM root component: %w", err)
-	} else if rootComponent == nil {
-		return nil, fmt.Errorf("No SBOM root component")
+	} 
+	if rootComponent != nil {
+		doc.Metadata.Component = rootComponent
 	}
-
-	doc.Metadata.Component = rootComponent
 	if err := s.componentsMaps(ctx, bom); err != nil {
 		return nil, err
 	}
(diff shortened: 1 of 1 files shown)


jspeed-meyers (approved):

puerco (approved):
Thanks @veramine! I think this needs to go in, but not for the reason you are proposing :) There is more than one scenario where we want to treat "rootless" CycloneDX components as multi-root sboms (see ). This means that we should be able to render them.

If we don't analyse the graph first, just rendering SBOMs without root nodes like this could potentially result in corruption of the graph as nodes may unexpectedly become roots (in the semantical sense of the produced multi-root CDX). It will work for now but we need to be aware that the produces SBOM may not be a good representation of the protobom when this happens.

Rootless sboms SBOMs are, in my view, invalid. Even when the spec allows you to build them. (we have discussed from spdx with Anchore BTW). I think protobom should fail when rendering a document without nodes...  __BUT__ (strong exception!) I do recognize that there are countless SBOMs out there structured like this. So my proposal is that we build a hack, triggered by an option:

When the option is set, if an SBOM has zero roots, we reconnect any lingering nodes as roots. Lingering nodes are those who are only "origins", that is nothing connects _to_ them. The very good news is that we already have logic built in that does this:

https://github.com/bom-squad/protobom/blob/597cac6b33c625d82db84e2cc2e1c00eed1be55e/pkg/sbom/nodelist.go#L614-L631
 
In order to keep the graph integral, we default to not using this in the library, but in any user facing tools like `sbom-convert`, as well as any apps which choose to do so, we can enable the hack via the relevant option.

So, in this goes but we should add that analysis to handle rootless sboms better. Thanks for working on this!

PR #133, opened 2023-10-17
Pointer bugs

puerco:
This PR fixes two bugs related to node comparison:

- It fixes bug that was indexing memory locations of string slices when generating the flattened string of the node.
- Fixes a bug where pointers where not being copied correctly in the `node.Copy()` function.


jspeed-meyers:
The tests are grumpy. 😢 Thoughts?

veramine (approved):
I tested this today and it works.  I added a new test to ensure Copy continues to work.  (https://github.com/bom-squad/protobom/pull/137)

Approved.

PR #139, opened 2023-10-19
ignore lint errors of type SA9003: empty branch

veramine:
Proposal to allow lint errors of type empty branch until we get to protobom v1.0.  

I want to add checks for degradation like this:

```golang
if len(node.PrimaryPurpose) > 1 {
	// TODO(degradation): Multiple PrimaryPurpose in protobom.Node, but spdx.Package only allows single PrimaryPackagePurpose so we are using the first
}
```

But I can't because lint error of empty branch.  So instead I do things like this:

```golang
if len(node.PrimaryPurpose) > 1 {
	// TODO(degradation): Multiple PrimaryPurpose in protobom.Node, but spdx.Package only allows single PrimaryPackagePurpose so we are using the first
	if true { // temp workaround in favor of adding a lint tag
		break
	}
}
```

and this

```golang
func noop() {}
...
default:
	// TODO(degradation): Non-matching primary purpose to component type mapping
	noop() // temp workaround in favor of adding a lint tag
```

I'm proposing we relax the lint checks for empty branches so we can leave these empty branches we intend to fill in prior to v1.0.  Here's the change:

```yml
    - linters:
        - staticcheck
      text: "^SA9003:"  # Ignore empty branch lint errors while we are scaffolding new functionality
```

Here's the kind of errors we would stop getting on check-in:

```
pkg/native/serializers/serializer_cdx.go:308:3: SA9003: empty branch (staticcheck)
		if len(n.PrimaryPurpose) > 1 {
```

### DIFF
--- a/.golangci.yaml
+++ b/.golangci.yaml
@@ -10,6 +10,9 @@ issues:
         - gocritic
         - golint
         - dupl
+    - linters:
+        - staticcheck
+      text: "^SA9003:"  # Ignore empty branch lint errors while we are scaffolding new functionality
 
   # Maximum issues count per one linter. Set to 0 to disable. Default is 50.
   max-issues-per-linter: 0


puerco (commented):
I think I would keep the linter, but if you want to disable it to iterate quickly _and_ if you pinky promise to re-enable it before we cut a new release, especially while we work on the degradation mechanism we could disable it temporarily. Keep in mind that in the end it could be more work 😉 

In general, the empty code paths are nothing but dead code, we could just remove them to keep  the linter happy:

For example, this example could be removed, just leaving the TODO comment:

```golang
if len(node.PrimaryPurpose) > 1 {
	// TODO(degradation): Multiple PrimaryPurpose in protobom.Node, but spdx.Package only allows single PrimaryPackagePurpose so we are using the first
}
```

While on this one, you can remove the `if true` block with the same effect:

```golang
if len(node.PrimaryPurpose) > 1 {
	// TODO(degradation): Multiple PrimaryPurpose in protobom.Node, but spdx.Package only allows single PrimaryPackagePurpose so we are using the first
	if true { // temp workaround in favor of adding a lint tag
		break
	}
}
```

Your call, I'd rather keep the linter as protobom is already going to some more serious applications.

veramine:
ehhh... I don't want to pinky promise...  I'll just use comments instead... I've gotten used to coding a certain way but I understand it's not just used by me... :)

PR #140, opened 2023-10-20
Graph API improvements

puerco:
This PR introduces new functions to the protobom graph API. It adds a few new capabilities to make life easier when working with SBOMs and new functionality.

New features:

-  `nodelist.NodeDescendants` is a new function that traverses the graph starting from a node and returns all linked nodes up to a set distance.
- `nodelist.AddRootNode` is the equivalent of `nodelist.AddNode` but it also registers the node as a root node.
- `nodelist.RelateNodeAtId` inserts a node into the graph by relating it to a specified node.

Quality of life improvements:

- `node.AddHash` is a convenience function to add hashes to nodes.
- We now have constructors for the protobom elements (eg `n := NewNode()` )

I've added tests for all new functions and simplified the readme with the new functions.

### DIFF
--- a/README.md
+++ b/README.md
@@ -110,7 +110,7 @@ func main() {
 	// Create a node to represent the application:
 	appNode := &sbom.Node{
 		Id:               "pkg:generic/my-software@v1.0.0",
-		PrimaryPurpose:   "application",
+		PrimaryPurpose:   []sbom.Purpose{sbom.Purpose_APPLICATION},
 		Name:             "My Software Name",
 		Version:          "v1.0.0",
 		Licenses:         []string{"Apache-2.0"},
--- a/README.md
+++ b/README.md
@@ -119,9 +119,7 @@ func main() {
 	}
 
 	// Add the application node to the document's nodelist:
-	document.NodeList.AddNode(appNode)
-	// ... and to the document's root elements:
-	document.NodeList.RootElements = append(document.NodeList.RootElements, appNode.Id)
+	document.NodeList.AddRootNode(appNode)
 
 	// Create two nodes to describe files in the application
 
@@ -134,11 +132,9 @@ func main() {
 		Description: "Software Lib",
 	}
 
-	node1.Hashes = map[int32]string{
-		int32(sbom.HashAlgorithm_SHA1):   "f3ae11065cafc14e27a1410ae8be28e600bb8336",
-		int32(sbom.HashAlgorithm_SHA256): "4f232eeb99e1663d07f0af1af6ea262bf594934b694228e71fd8f159f9a19f32",
-		int32(sbom.HashAlgorithm_SHA512): "8044d0df34242699ad73bfe99b9ac3d6bbdaa4f8ebce1e23ee5c7f9fe59db8ad7b01fe94e886941793aee802008a35b05a30bc51426db796aa21e5e91b7ed9be",
-	}
+	node1.AddHash(sbom.HashAlgorithm_SHA1, "f3ae11065cafc14e27a1410ae8be28e600bb8336")
+	node1.AddHash(sbom.HashAlgorithm_SHA256, "4f232eeb99e1663d07f0af1af6ea262bf594934b694228e71fd8f159f9a19f32")
+	node1.AddHash(sbom.HashAlgorithm_SHA512, "8044d0df34242699ad73bfe99b9ac3d6bbdaa4f8ebce1e23ee5c7f9fe59db8ad7b01fe94e886941793aee802008a35b05a30bc51426db796aa21e5e91b7ed9be")
 
 	document.NodeList.AddNode(node1)
 
@@ -146,27 +142,19 @@ func main() {
 		Id:          "File--usr-bin-software",
 		Type:        sbom.Node_FILE,
 		Name:        "/usr/bin/software",
-		Version:     "1",
+		Version:     "1.0",
 		Copyright:   "Copyright 2023 The ACME Corporation",
 		Description: "Software binary",
 	}
 
-	node2.Hashes = map[int32]string{
-		int32(sbom.HashAlgorithm_SHA1):   "defee82004d22fc92ab81c0c952a62a2172bda8c",
-		int32(sbom.HashAlgorithm_SHA256): "ad291c9572af8fc2ec8fd78d295adf7132c60ad3d10488fb63d120fc967a4132",
-		int32(sbom.HashAlgorithm_SHA512): "5940d8647907831e77ec00d81b318ca06655dbb0fd36d112684b03947412f0f98ea85b32548bc0877f3d7ce8f4de9b2c964062df44742b98c8e9bd851faecce9",
-	}
+	node2.AddHash(sbom.HashAlgorithm_SHA1, "defee82004d22fc92ab81c0c952a62a2172bda8c")
+	node2.AddHash(sbom.HashAlgorithm_SHA256, "ad291c9572af8fc2ec8fd78d295adf7132c60ad3d10488fb63d120fc967a4132")
+	node2.AddHash(sbom.HashAlgorithm_SHA512,  "5940d8647907831e77ec00d81b318ca06655dbb0fd36d112684b03947412f0f98ea85b32548bc0877f3d7ce8f4de9b2c964062df44742b98c8e9bd851faecce9")
 
-	document.NodeList.AddNode(node2)
-
-	// Relate the application package and the files:
-	edge := &sbom.Edge{
-		Type: sbom.Edge_contains,
-		From: appNode.Id,
-		To:   []string{node1.Id, node2.Id},
-	}
-
-	document.NodeList.Edges = append(document.NodeList.Edges, edge)
+	// Relate the application package and the files. This adds the nodes to 
+	// the document graph:
+	document.RelateNodeAtID(node1, appNode.Id, sbom.Edge_contains)
+	document.RelateNodeAtID(node2, appNode.Id, sbom.Edge_contains)
 
 	// Now render the document to STDOUT:
 	w := writer.New()
--- a/pkg/sbom/edge.go
+++ b/pkg/sbom/edge.go
@@ -5,6 +5,12 @@ import (
 	"strings"
 )
 
+func NewEdge() *Edge {
+	return &Edge{
+		To: []string{},
+	}
+}
+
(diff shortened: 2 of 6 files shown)


pkg/sbom/node_test.go
```diff
@@ -304,3 +304,152 @@ func TestNodeCopy(t *testing.T) {
...
+			//     mynode
+			//       |
+			name: "single node",
+			sut: &NodeList{
+				Nodes:        []*Node{{Id: sutId}},
+				Edges:        []*Edge{},
+				RootElements: []string{sutId},
+			},
+			expectedNodesLength: 1,
+			depth:               10,
```
jspeed-meyers, line 326: I don't understand why depth is expected to be 10 here? I'm definitely missing something.

jspeed-meyers (approved):
Just a question (for myself really, and tiny typo)

README.md
```diff
@@ -134,39 +132,29 @@ func main() {
...
 	}
 
-	node1.Hashes = map[int32]string{
-		int32(sbom.HashAlgorithm_SHA1):   "f3ae11065cafc14e27a1410ae8be28e600bb8336",
-		int32(sbom.HashAlgorithm_SHA256): "4f232eeb99e1663d07f0af1af6ea262bf594934b694228e71fd8f159f9a19f32",
-		int32(sbom.HashAlgorithm_SHA512): "8044d0df34242699ad73bfe99b9ac3d6bbdaa4f8ebce1e23ee5c7f9fe59db8ad7b01fe94e886941793aee802008a35b05a30bc51426db796aa21e5e91b7ed9be",
-	}
+	node1.AddHash(sbom.HashAlgorithm_SHA1, "f3ae11065cafc14e27a1410ae8be28e600bb8336")
+	node1.AddHash(sbom.HashAlgorithm_SHA256, "4f232eeb99e1663d07f0af1af6ea262bf594934b694228e71fd8f159f9a19f32")
+	node1.AddHash(sbom.HashAlgorithm_SHA512, "8044d0df34242699ad73bfe99b9ac3d6bbdaa4f8ebce1e23ee5c7f9fe59db8ad7b01fe94e886941793aee802008a35b05a30bc51426db796aa21e5e91b7ed9be")
```
veramine, line 137: At this moment, node1.AddHash causes panic because Hashes is not allocated in this example.  We should check for n.Hashes being nil in AddHash().  I'll suggest that below in node.go

README.md
```diff
@@ -134,39 +132,29 @@ func main() {
...
-
-	// Relate the application package and the files:
-	edge := &sbom.Edge{
-		Type: sbom.Edge_contains,
-		From: appNode.Id,
-		To:   []string{node1.Id, node2.Id},
-	}
+	node2.AddHash(sbom.HashAlgorithm_SHA1, "defee82004d22fc92ab81c0c952a62a2172bda8c")
+	node2.AddHash(sbom.HashAlgorithm_SHA256, "ad291c9572af8fc2ec8fd78d295adf7132c60ad3d10488fb63d120fc967a4132")
+	node2.AddHash(sbom.HashAlgorithm_SHA512,  "5940d8647907831e77ec00d81b318ca06655dbb0fd36d112684b03947412f0f98ea85b32548bc0877f3d7ce8f4de9b2c964062df44742b98c8e9bd851faecce9")
```
veramine, line 152: (same here)

README.md
```diff
@@ -134,39 +132,29 @@ func main() {
...
 
-	document.NodeList.Edges = append(document.NodeList.Edges, edge)
+	// Relate the application package and the files. This adds the nodes to 
+	// the document graph:
+	document.RelateNodeAtID(node1, appNode.Id, sbom.Edge_contains)
+	document.RelateNodeAtID(node2, appNode.Id, sbom.Edge_contains)
```
veramine, line 157: I love this function idea, yes!

pkg/sbom/node.go
```diff
@@ -16,6 +16,19 @@ import (
...
+		Attribution:        []string{},
+		Suppliers:          []*Person{},
+		Originators:        []*Person{},
+		ExternalReferences: []*ExternalReference{},
+		FileTypes:          []string{},
+		Identifiers:        map[int32]string{},
+		Hashes:             map[int32]string{},
+	}
+}
+
```
veramine, line 31: if people use NewNode(), Hashes is populated.  But we should assume some people (like me) will just happily 

```
	node1 := &sbom.Node{
                SomeField: SomeValue
        }
```

And we should protect against that before trying to use or set any values that assume it has been previously created with the NewNode() constructor

pkg/sbom/node.go
```diff
@@ -370,3 +383,12 @@ func (n *Node) HashesMatch(th map[int32]string) bool {
 	}
 	return atLeastOneMatch
 }
+
+// AddHash adds a new hash of algorithm algo to the node. If the node
+// already has a has of the same algorithm it will get silently replaced.
```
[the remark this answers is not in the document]
veramine: I've read that code three times this morning already and never saw the typo lol

pkg/sbom/node.go
```diff
@@ -370,3 +383,12 @@ func (n *Node) HashesMatch(th map[int32]string) bool {
...
+func (n *Node) AddHash(algo HashAlgorithm, value string) {
+	if value == "" {
+		return
+	}
+	n.Hashes[int32(algo)] = value
```
veramine, line 393: Howabout 
```
	if value == "" {
		return
	}
        if n.Hashes == nil {
		n.Hashes = map[int32]string{}
	}
        n.Hashes[int32(algo)] = value
```

README.md
veramine: Or, the example could use NewNode().  But I still think node.go AddHash() should protect us against nil Hashes.

pkg/sbom/nodelist.go
```diff
@@ -749,3 +817,78 @@ func (nl *NodeList) NodeSiblings(id string) *NodeList {
...
+	}
+
+	siblings := nodeIndex{}
+
+	var loopNodes []*Node
+	newLoopNodes := []*Node{}
+
+	for i := 0; i < maxDepth; i++ {
+		if i == 0 {
+			fmt.Fprintf(os.Stdout, "Nuevo\n")
```
veramine, line 846: I get happy little 

```
Nuevo
Nuevo
Nuevo
Nuevo
Nuevo
```

output from go test when I run it now.  Probably don't need this :)

pkg/sbom/nodelist.go
```diff
@@ -539,6 +567,46 @@ func (nl *NodeList) Equal(nl2 *NodeList) bool {
 	return cmp.Equal(nlNodes, nl2Nodes)
 }
 
+// RelateNodeAtID creates a relationship between Node n and an existing node
+// in the NodeList specified by nodeID. If the node (as looked up by ID) does not
+// not exist in the NodeList it will be added. If NodeID does not exist an error
+// will be returned.
+func (nl *NodeList) RelateNodeAtID(n *Node, nodeID string, edgeType Edge_Type) error {
```
veramine, line 574: very nice function, thank you!

README.md
```diff
@@ -119,9 +119,7 @@ func main() {
 	}
```
veramine, line 119: Not your fault, but we also need to change the README example code PrimaryPurpose definition to be a slice or this example won't build.

Suggestion:
```
PrimaryPurpose:   []sbom.Purpose{sbom.Purpose_APPLICATION},
```

pkg/sbom/edge.go
```diff
@@ -5,6 +5,12 @@ import (
 	"strings"
 )
 
+func NewEdge() *Edge {
+	return &Edge{
+		To: []string{},
+	}
+}
+
```
veramine, line 13: I see "yay new function!" and then immediately think "ohh... need new test..." lol. I'll add edge_test.go after this is merged.

veramine (changes requested):
Very nice additions, thank you!  Need to make sure the README example builds cleanly and AddHash doesn't panic.  Otherwise, wonderful changes.

pkg/sbom/nodelist.go
puerco: Ah sorry I left this debugger string. Thanks, fixed!

pkg/sbom/node.go
puerco: Very nice catch. I added your suggestion to the commit

pkg/sbom/node.go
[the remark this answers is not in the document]
puerco: Fixed thanks :)

README.md
puerco: Yeah this is why i missed it, I was testing only with nodes built with NewNode(). Thanks!

pkg/sbom/node.go
puerco: 100%

README.md
puerco: Sorry for not testing the example before submitting it. It as really a last minute improvement I did to the PR. At least it helped us catch a bug :). It runs now:

```bash
go run test.go
```

```json
{
    "spdxVersion": "SPDX-2.3",
    "dataLicense": "CC0-1.0",
    "SPDXID": "SPDXRef-DOCUMENT",
    "name": "",
    "documentNamespace": "https://spdx.org/spdxdocs/",
    "creationInfo": {
        "licenseListVersion": "3.20",
        "creators": [
            "Tool: protobom-devel",
            "Tool: ACME SBOM Tool-1.0"
        ],
        "created": "2023-10-21T00:16:25Z"
    },
    "packages": [
        {
            "name": "My Software Name",
            "SPDXID": "SPDXRef-pkg:generic/my-software@v1.0.0",
            "versionInfo": "v1.0.0",
            "downloadLocation": "NOASSERTION",
            "filesAnalyzed": false,
            "licenseConcluded": "Apache-2.0",
            "licenseComments": "Apache License",
            "primaryPackagePurpose": "APPLICATION"
        }

```

puerco:
Thanks for the thorough review @veramine and @jspeed-meyers , it should be ready to go now!

pkg/sbom/node_test.go
puerco: Ah this is just a "big enough number" to capture the expected (1-node) graph. It is larger than than the expected length because we want to catch any bugs in the traversal that may be causing duplicates.

veramine (approved):
Look great, thanks @puerco !

PR #141, opened 2023-10-20
Node diffing!

puerco:
This PR is the first of a series to build diffing capabilities into the protobom core. This PR starts with diffing support in the node. To compare two nodes, you simply call the new node.Diff method and pass it a second node to compare:

```golang
diff := node1.Diff(node2)
```

The result is a new simple structure called `NodeDiff` that captures in two nodes data from fields that were changed or added, data that was removed and the number of changes:

```golang
type NodeDiff struct {
	Added     *Node
	Removed   *Node
	DiffCount int
}
```

I've added some tests to the diff function that verify the integration with some of the helper functions, they can be improved and we should probably create a conformance framework to test diffing.

Note: Needs #140 to run

### DIFF
--- a/pkg/sbom/diff.go
+++ b/pkg/sbom/diff.go
@@ -1,0 +1,146 @@
+package sbom
+
+import (
+	"slices"
+
+	timestamppb "google.golang.org/protobuf/types/known/timestamppb"
+)
+
+// diffString takes compares s2 against s1. If they differ returns the value of
+// s2 in the removed return value. If s2 is blank, removed will have the original
+// value of s1. If the strings are differente count will be 1, zero if not.
+func diffString(s1, s2 string) (added, removed string, count int) {
+	if s1 == s2 {
+		return "", "", 0
+	}
+
+	if s2 == "" {
+		return "", s1, 1
+	}
+	return s2, "", 1
+}
+
+// diffDates takes two dates, comapres them and returns d2 in added if there is
+// a change, s1 in removed if d2 is nil. count will be 1 if there was a change.
+func diffDates(d1, d2 *timestamppb.Timestamp) (added, removed *timestamppb.Timestamp, count int) {
+	if (d1 != nil && d2 != nil && d1 != d2) || (d1 == nil && d2 != nil) {
+		return d2, nil, 1
+	} else if d1 != nil && d2 == nil {
+		return nil, d1, 1
+	}
+	return nil, nil, 0
+}
+
+func diffPersonList(list1, list2 []*Person) (added, removed []*Person, count int) {
+	added = []*Person{}
+	removed = []*Person{}
+
+	// Index persons
+	idx1 := map[string]*Person{}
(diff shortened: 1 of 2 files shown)


jspeed-meyers:
@puerco: `golangci-lint` is grumpy about a few minor issues. After you fix that up, I'm glad to put in a review.

This is awesome!

puerco:
OK, I've fixed the linter and rebased it to head, PTAL!

pkg/sbom/diff.go
```diff
@@ -0,0 +1,329 @@
...
+
+type NodeDiff struct {
+	Added     *Node
+	Removed   *Node
+	DiffCount int
+}
+
+// Diff analyses a node and returns a a new node populated with all fields
+// that are different in n2 from n. If no changes are found, Diff returns nil
+func (n *Node) Diff(n2 *Node) *NodeDiff {
```
jspeed-meyers, line 18: Whoa, this is cool!

pkg/sbom/diff.go
```diff
@@ -0,0 +1,329 @@
...
+	nd.Removed.Hashes = removedM
+	nd.DiffCount += count
+
+	if nd.DiffCount > 0 {
+		return &nd
+	}
+	return nil
+}
+
+// diffString takes compares s2 against s1. If they differ returns the value of
```
jspeed-meyers, line 160:
```suggestion
// diffString compares s2 against s1. If they differ returns the value of
```

jspeed-meyers (approved):

veramine (approved):
Oh my goodness, this is beautiful!  Thank you for your contributions.  I know this kind of work is tedious.  Very much appreciate how comprehensive this effort was!!  

Only thing that took me a second to realize was the function signature from the differs..  (added, removed string, count int).  When you pass in string "ABC" and string "DEF", the "DEF" is reported as "added".  I'm not sure "added" is the right word though, right?  It's just a different value...  But I'm still happy with what you've done, I think its great.  It was just the first thought I had while reading.  And I looked for any missing loopholes and couldn't find anything, good job!!

puerco:
@veramine I know what you mean. I tried to think of a return structure we could generalize to all data types to keep just one mental model. My inspiration the simplest example I could think of:  a diff (as in a patch) where what you get is a report of how data looked like then and how it looks now. 

I'll merge it for now but I'm open to changing the return struct or the signatures to something else, especially before we cut the next release (I'm planning to cut it next Friday) :)

veramine:
I was thinking of diff as in the command line utility that shows

< original value
--
> new value

If it made sense, you could go in that way by returning a structure populated for each diff between the two with a "left" and "right".  Left being the < part of the diff.  Right being the > part of the diff.  And then use a structure something like:

type NodeDiffs struct {
	Left    *Node
	Right   *Node
	DiffCount int
}

Each diff is reported in both Left and Right.  

But your way is also good.  Whatever makes sense to you.

PR #146, opened 2023-11-27
Improve conformance tests and samples

puerco:
This PR expands the conformance tests we do in protobom to also ensure we are serializing/unserializing the edges, document and the nodelist.

It also adds a number of samples to the golden SPDX sample collection.

**Note:**

I ripped out the CycloneDX samples originally meant for this PR into  #147 as the conformance tests are catching bugs in the (un)serializers. I'll fix those and then we can merge the CDX samples.


puerco:
@veramine yes! I added warnings to the conformance generator to highlight whenever we are missing samples for a format.

veramine:
alright cool, this is fine with me then.  Thanks!

Issue #148, opened 2023-12-07
External reference type should be enumerated


puerco:
In our current model, the external reference type is a string:

https://github.com/bom-squad/protobom/blob/3dff6b2828bc60a260f8f8ddfdb7389b66844d1b/api/sbom.proto#L130-L139

Given that both formats enumerate the valid values we should also have an enumerated field.

veramine:
I asked chatgpt for list of valid values and got these:

SPDX:

* cpe22Type: This reference type is used for specifying a CPE (Common Platform Enumeration) version 2.2 identifier.
* cpe23Type: Similar to cpe22Type, but for CPE version 2.3 identifiers.
* maven-central: This reference is used for artifacts that are stored in Maven Central, typically for Java libraries.
* npm: This reference type is for packages that are available in the npm (Node Package Manager) registry.
* nuget: Refers to packages available in the NuGet gallery, commonly used for .NET packages.
* bower: For components that are available in the Bower registry, a package manager for front-end web development.
* purl: Package URLs (purl) provide a standard way to identify and locate software packages across different package managers and registries.
* swidTagID: Refers to the SWID (Software Identification) Tag ID as per ISO/IEC 19770-2:2015 standards.
* git: This reference is for Git repositories, specifying a particular commit, branch, or tag.
* other: A catch-all category for references that don't fit into the other types, allowing for flexibility and future expansion.

CycloneDX:

* vcs: Version Control System reference. It's used to point to the source code repository of the component.
* issue-tracker: Reference to the issue tracker associated with the component (e.g., JIRA, GitHub issues).
* website: A reference to the official website or webpage for the component.
* advisories: Used to link to published advisories related to the component.
* bom: This type of reference points to another SBOM that is related to the component.
* mailing-list: Reference to a mailing list that is relevant to the component.
* social: Refers to social media or other online platforms where discussions or information about the component can be found.
* chat: Reference to a chat platform or channel related to the component (e.g., Slack channels, IRC).
* documentation: Links to the official documentation of the component.
* support: Reference to support resources for the component, such as help desks or support forums.
* distribution: Used to reference a distribution point for the component, like a package registry or download page.
* license: A reference to the license details of the component.
* build-meta: Reference to build metadata.
* build-system: Points to the system or tool used for building the component.
* other: A catch-all category for references that don't fit into the other types, similar to SPDX.

Do these seem right?  If so, I can handle this issue.

puerco:
[...]

PR #149, opened 2023-12-08
Enumerate External Reference Types

puerco:
⚠️  **This is a breaking change** ⚠️ 

As discussed in the cohort meeting of Dec 6th, the external reference types should be enumerated instead of a free-form string. This PR addresses it

Fixes https://github.com/bom-squad/protobom/issues/148

This PR also removes some of the legacy functions from the `ExternalReference` object that now are incompatible and to better conform to the design principle that dictates that format-specific logic should live only in the serializers/unserializers.

I added new tests to cover the functionality that was moved into the native objects.

The mapping of the reference types can be looked up in this spreadsheet. I will document in markdown in a follow up PR.


jspeed-meyers (approved):
Nice work, @puerco! LGTM

pkg/native/serializers/serializer_spdx23.go
```diff
@@ -353,3 +357,53 @@ func buildPackages(bom *sbom.Document) ([]*spdx.Package, error) { //nolint:unpar
...
 }
+
+// ExtRefCategoryFromProtobomExtRef reads a protobom external reference struct and returns a
+// string with the corresponding category
+func (s *SPDX23) extRefCategoryFromProtobomExtRef(extref *sbom.ExternalReference) string {
+	switch extref.Type {
+	case sbom.ExternalReference_BOWER, sbom.ExternalReference_MAVEN_CENTRAL,
+		sbom.ExternalReference_NPM, sbom.ExternalReference_NUGET:
+		return spdx.CategoryPackageManager
+	case sbom.ExternalReference_SECURITY_ADVISORY, sbom.ExternalReference_SECURITY_FIX:
```
veramine, line 368: what about adding sbom.ExternalReference_SWID here?

api/sbom.proto
```diff
@@ -130,12 +130,79 @@ message Edge {
...
+        RELEASE_HISTORY = 38;
+        RELEASE_NOTES = 39;
+        RISK_ASSESSMENT = 40;
+        RUNTIME_ANALYSIS_REPORT = 41;
+        SECURE_SOFTWARE_ATTESTATION = 42;
+        SECURITY_ADVERSARY_MODEL = 43;
+        SECURITY_ADVISORY = 44;
+        SECURITY_CONTACT = 45;
+        SECURITY_FIX = 46;
+        SECURITY_OTHER = 47;
```
veramine, line 191: Is this just url right now?  Are you thinking of using SECURITY_OTHER for more than just url?  If its just url, you could just call it SECURITY_URL instead of SECURITY_OTHER..  That would prevent us from having "OTHER" and SECURITY_OTHER.

veramine (commented):
I've just reviewed the lists so far.  I haven't tested myself yet.  I will clone your repo and test for myself tomorrow morning.  I added a couple comments about security_other and swid above.  I also wonder if we should include support for cpe instead we're doing this anyway now.  And what about cdx?  Not to diminish the work you've done here on spdx, it's fantastic.  But maybe we should just get it all done in one PR if you can stomach the copy-pasting...

api/sbom.proto
puerco: SPDX3 has a security other, we can store the 2.x security+url here. None of the formats we target has a "URL" type (except the security+URL extref in spdx 2)

pkg/native/serializers/serializer_spdx23.go
puerco: Oh yeah. good catch. Can I push it in a follow-up PR to avoid rebase hell in the next PRs?

veramine: Of course!

api/sbom.proto
veramine: alright cool

puerco:
> And what about cdx? 

I just discovered that external references are not implemented in cdx, I'm about to open a PR for that :+1:

veramine (approved):
Totally fine with me to go ahead with what you have now and keep working on it.  Thanks!

puerco:
Allright, thanks this helps a lot. Let me know if you find something wrong in your tests. Thank you!

PR #147, opened 2023-11-27
Add CycloneDX conformance samples

puerco:
This PR adds couple of CycloneDX documents to the golden samples collection of our conformance tests.

 Signed-off-by: Adolfo Garcia Veytia (puerco) <puerco@chainguard.dev>

PR #151, opened 2023-12-09
CycloneDX Fixes

puerco:
This PR adds various improvements and fixes to our CycloneDX code to ensure we can test conformance when working with CDX SBOMs.

There were a number of conformance problems uncovered when adding the first CycloneDX test samples (see https://github.com/bom-squad/protobom/pull/147) and this PR is the first set of changes to fix them,

There are three parts of this PR:

**Support for External References in CycloneDX**

We were lacking proper support for external references. This PR uses the new enumeration to implement extrefs.

**Refactor conversion logic to make it testable**

The conversion functions, that is those that convert between cyclonedx constants and our enumerated values, are now split out to their own utility functions which make it easier to reuse, to understand what they do and most importantly, now have tests to ensure they are working as expected,

**Deterministic Autogenerated IDs**

When we encounter a cdx component without a bom-ref, we generate an ID. We used to use random strings but it made it impossible to test conformance. The CDX unserializer will now generate deterministic IDs so that protobom documents read from cyclonedx data will now match our golden samples

This PR also includes the first set of cyclonedx conformance data. I'll keep adding more in subsequent PRs as I fix more bugs.


pkg/native/serializers/serializer_cdx.go
```diff
@@ -46,27 +46,47 @@ func (s *CDX) Serialize(bom *sbom.Document, _ *native.SerializeOptions, _ interf
 	doc := cdx.NewBOM()
 	doc.SerialNumber = bom.Metadata.Id
 	ver, err := strconv.Atoi(bom.Metadata.Version)
+	// TODO(deprecation): If varsion does not parse to int, data loos
```
jspeed-meyers, line 49:
```suggestion
	// TODO(deprecation): If version does not parse to int, data loss
```

pkg/native/serializers/serializer_cdx.go
```diff
@@ -491,3 +467,161 @@ func getCDXState(ctx context.Context) (*serializerCDXState, error) {
 	}
 	return dm, nil
 }
+
+// protobomExtRefTypeToCdxType translates between the protobom external reference
+// identiers and the cycloneDX equivalent types.
```
jspeed-meyers, line 472:
```suggestion
// identifiers and the cycloneDX equivalent types.
```

pkg/native/serializers/serializer_cdx.go
```diff
@@ -491,3 +467,161 @@ func getCDXState(ctx context.Context) (*serializerCDXState, error) {
...
+	case sbom.HashAlgorithm_BLAKE3:
+		return cdx.HashAlgoBlake3, nil
+	}
+
+	// TODO(degradation): Unknow algorithms err here. We could silently not.
+	// TODO(options): Sink all unknows to UNKNOWN
+	return "", fmt.Errorf("hash algorithm %q not supported by cyclonedx", protoAlgo)
+}
+
+// purposeToComponentType convertos from a protobom enumarted purpose to
```
jspeed-meyers, line 596:
```suggestion
// purposeToComponentType converts from a protobom enumerated purpose to
```

jspeed-meyers (approved):
Nice. Squash those bugs! 🐛

puerco:
OK, I' merging this to continue on the CDX updates,

PR #155, opened 2023-12-23
Bump CDX libraries to 0.8

puerco:
This PR updates the cycloneDX libraries to v0.8.0. It only fixes the serializer to use the new ToolChoice struct introduced in the latest version. It does not yet support the new component-based tool.

/cc @manifestori @veramine 


manifestori (approved):
LGTM

PR #160, opened 2023-12-26
Add Release Workflow

puerco:
Add a release workflow to automate the release cut on tag push


jspeed-meyers (approved):
YAY. Thanks, @puerco!

PR #161, opened 2023-12-28
tmp diff test

houdini91:
This PR is temporary playground of a diff test i wanted to pass around.

I suggest a test as following:
1) Read some format (`from`) file in to a protobom document.
1) Translate protobm in to a other format (`to`) 
2) Read back the translated (`to`) document back in to the protobom document.
2) Option 1: Compare the original document protobom (`from`) to the translated one (`to`).
2) Option 2: Compare the original (`from`) golden document the translated one (`to`)
2) Optoin 3: Create a NEW translated **golden** document and compare it to the translated one (`to`)

What we can test,
We can start follow what fields are persistent across translation and what are not.
We don't have to have a perfect score on the diff just yet , but i can be a starting point to solidify what we expect to change across a translation and what we don't.

Open issues: 
* Test does not actually pass there seems there are differences we yet to formalize in to the test.
* A detailed investigation of all losing info.
PS: PR is far from ready i just wanted to share code to scketch this out.

### DIFF
--- a/test/conformance/conformance_diff_test.go
+++ b/test/conformance/conformance_diff_test.go
@@ -1,0 +1,131 @@
+package conformance
+
+import (
+	"bufio"
+	"bytes"
+	"fmt"
+	"reflect"
+	"sort"
+	"testing"
+
+	"github.com/bom-squad/protobom/pkg/formats"
+	"github.com/bom-squad/protobom/pkg/reader"
+	"github.com/bom-squad/protobom/pkg/sbom"
+	"github.com/bom-squad/protobom/pkg/writer"
+	"github.com/google/go-cmp/cmp"
+	"github.com/stretchr/testify/require"
+)
+
+type fakeBuf struct {
+	*bufio.ReadWriter
+}
+
+func (fwc *fakeBuf) Close() error {
+	return nil
+}
+
+func (f *fakeBuf) Seek(offset int64, whence int) (int64, error) {
+	return 0, nil
+}
+
+func TestDiffTranslationFormats(t *testing.T) {
+	from := formats.SPDX23JSON
+	to := formats.CDX14JSON
+
+	fromFiles := findFiles(t, from)
+	require.NotEqual(t, len(fromFiles), 0)
+
+	fname := fromFiles[0]
+
+	r := reader.New()
+	fromSut, err := r.ParseFile(fname)
+
+	require.NoError(t, err)
+	w := writer.New()
+
+	var b bytes.Buffer
+	fakeWriter := bufio.NewWriter(&b)
+	fakeReader := bufio.NewReader(&b)
+	fakeReadWriter := bufio.NewReadWriter(fakeReader, fakeWriter)
+
+	fake := fakeBuf{
+		ReadWriter: fakeReadWriter,
+	}
+
+	w.WriteStreamWithOptions(
+		fromSut, &fake, &writer.Options{Format: to},
+	)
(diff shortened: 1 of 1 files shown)


test/conformance/conformance_diff_test.go
```diff
@@ -0,0 +1,131 @@
...
+		ReadWriter: fakeReadWriter,
+	}
+
+	w.WriteStreamWithOptions(
+		fromSut, &fake, &writer.Options{Format: to},
+	)
+	toSut, err := r.ParseStreamWithOptions(&fake, &reader.Options{Format: to})
+	require.NoError(t, err)
+
+	golden := readProtobom(t, fname+".proto")
```
houdini91, line 61: we can use fromSut but for the sake for conformance i chose to use golden proto.

test/conformance/conformance_diff_test.go
```diff
@@ -0,0 +1,131 @@
...
+}
+
+func (f *fakeBuf) Seek(offset int64, whence int) (int64, error) {
+	return 0, nil
+}
+
+func TestDiffTranslationFormats(t *testing.T) {
+	from := formats.SPDX23JSON
+	to := formats.CDX14JSON
+
```
puerco, line 34: Remember that we don't care about going from one format to another. protobom restricts its opinions to/from the native formats to the neutral representation.

test/conformance/conformance_diff_test.go
```diff
@@ -0,0 +1,131 @@
...
+	}
+
+	if len(nl.RootElements) != len(nl2.RootElements) {
+		require.Equal(t, len(nl.RootElements), len(nl2.RootElements), "root amount differs")
+	}
+
+	r1 := nl.RootElements
+	r2 := nl2.RootElements
+	sort.Strings(r1)
+	sort.Strings(r2)
```
puerco, line 94: You are on the right track here, but these expanded tests should augment the ones we have in the conformance tests which are very simple right now:

https://github.com/bom-squad/protobom/blob/c2d38e911b0b468130d2b6257e219ff84ef7d689/test/conformance/conformance_test.go#L60-L63

We should start adding more deeper test cases. Right now what we do is simply test that the documents more or less "look the same", that is have the same number of nodes and edges, etc. But we should add more comprehensive comparisons.

Also remember that we already have diff support in the protobom API, so avoid duplicating that.

puerco (commented):
OK, this is pretty cool @houdini91 just but remember that we need to ensure conformance to/from the native formats to the neutral representation. Not from format to format to curve complexity.

We need some of the expanded tests you are writing to live in the conformance suite, so just integrate them there, I added a note below :point_down:

test/conformance/conformance_diff_test.go
houdini91: Yeah i tried to use the simple string compare failed to match when running the test .. Let me review the diff probably missed it..

houdini91:
@puerco I not sure i fully understood your comment.

> just but remember that we need to ensure conformance to/from the native formats to the neutral representation. Not from format to format to curve complexity.

Do you mean like the current conformance tests logic where there is one reading operation in to the neutral representation or similar to my suggestion where there is another writer operations (then another read to return to the neutral representation?).

test/conformance/conformance_diff_test.go
houdini91: I decided to expand the current conformance on a separate PR so this one can stay about diff between the source and translated version of the same document.

Take a peek in https://github.com/bom-squad/protobom/pull/163 see if correlates.

houdini91:
Up

PR #163, opened 2024-01-07
Conformance enhancement suggestion

houdini91:
* Small expansion of the testNode conformance function to reveal diff details.


houdini91:
A more advanced Diff func i was playing with was
```
// NodeListDiff represents the difference between two lists of nodes
type NodeListDiff struct {
	Added    []*sbom.Node
	Removed  []*sbom.Node
	NodeDiff []*sbom.NodeDiff
}

// DiffNodes compares two lists of nodes and returns the differences
func nodeListDiff(nl, nl2 []*sbom.Node) NodeListDiff {
	diff := NodeListDiff{}
	// Sort both node lists
	sort.Slice(nl, func(i, j int) bool {
		return strings.Compare(nl[i].String(), nl[j].String()) <= 0
	})
	sort.Slice(nl2, func(i, j int) bool {
		return strings.Compare(nl2[i].String(), nl2[j].String()) <= 0
	})

	index1, index2 := 0, 0

	// Iterate through both sorted node lists
	for index1 < len(nl) || index2 < len(nl2) {
		if index1 < len(nl) && index2 < len(nl2) {
			n, n2 := nl[index1], nl2[index2]
			switch strings.Compare(n.Id, n2.Id) { //Use ID to decide if to compare
			case 0: // Nodes are equal
				index1++
				index2++
				nodeDiff := n.Diff(n2)
				if nodeDiff != nil {
					diff.NodeDiff = append(diff.NodeDiff, nodeDiff)
				}
			case -1: // Node1 is less than Node2
				diff.Removed = append(diff.Removed, n)
				index1++
			case 1: // Node1 is greater than Node2
				diff.Added = append(diff.Added, n2)
				index2++
			}
		} else if index1 < len(nl) {
			diff.Removed = append(diff.Removed, nl[index1])
			index1++
		} else {
			diff.Added = append(diff.Added, nl2[index2])
			index2++
		}
	}

	return diff
}
```
@puerco What do you think?

houdini91:
@puerco  I opened a separate PR for NodeList.Diff suggestion (Currently Draft) https://github.com/bom-squad/protobom/pull/173

PR #173, opened 2024-01-18
NodeList Diff API suggestion

houdini91:
Suggestion a new API function, NodeList.Diff, is the primary focus of this pull request. I've provided a preliminary code sketch for your review and feedback.

**New API Function:** NodeList.Diff

**Code Sketch for Review:**
I have outlined the initial implementation of the NodeList.Diff function. Your valuable input on the code is highly appreciated.

**Scope:**
At this stage, the implementation concentrates on the NodeList object, specifically addressing nodes, edges and root elements.


houdini91:
Added edge diff support (still missing root diff)

houdini91:
Added rootElements diff support.

pkg/sbom/diff_nodelist.go
```diff
@@ -0,0 +1,256 @@
...
+			diff.Removed = append(diff.Removed, nlEdges[index1])
+			index1++
+		} else {
+			diff.Added = append(diff.Added, nl2Edges[index2])
+			index2++
+		}
+	}
+
+	sort.Slice(diff.Added, func(i, j int) bool {
+		return compareEdges(diff.Added[i], diff.Added[j]) <= 0
```
puerco, line 243: do this comparisons with the built-in Equal function (see comment below)

pkg/sbom/diff_nodelist.go
```diff
@@ -0,0 +1,256 @@
...
+	})
+
+	return diff
+}
+
+func compareEdges(e1, e2 *Edge) int {
+	sort.Strings(e1.To)
+	sort.Strings(e2.To)
+	return strings.Compare(e1.String(), e2.String())
+}
```
puerco, line 256: Drop this function and use `Edge.Equal(edge2)`. The protobuf string function pulls all data in the object including some that (in the future) we may not consider semantically valid for comparing.

pkg/sbom/diff_nodelist.go
```diff
@@ -0,0 +1,256 @@
...
+		EdgesDiff:         nl.diffEdges(nl2),
+		RootElmementsDiff: nl.diffRootElements(nl2),
+	}
+}
+
+// DiffWithStripQulifiers analyses a NodeList and returns a NodeList populated with all fields
+// that are different in nl2 from nl after stripping identifier Quilifiers.
+func (nl *NodeList) DiffWithStripQulifiers(nl2 *NodeList, strip bool) NodeListDiff {
+	return NodeListDiff{
+		NodesDiff:         nl.diffNodes(nl2, strip),
```
puerco, line 83: Wiring a stip parameter here will not scale well if we want to do other comparisons like this.

Instead, we should pass transformer functions that mutate the elements, I would suggest leaving the qualifier stripping for a follow-up pr to keep this PR easier to review.

pkg/sbom/diff_nodelist.go
```diff
@@ -0,0 +1,256 @@
...
+
+func stripQualifiers(url string) string {
+	// Define a regular expression pattern to match everything after "?"
+	pattern := regexp.MustCompile(`\?.*`)
+	// Substitute the matched pattern with an empty string to remove it
+	strippedURL := pattern.ReplaceAllString(url, "")
+	return strippedURL
+}
+
+func stripNodeQualifiers(n *Node) {
```
puerco, line 129: Watch out: This function is mutating the original node's properties as it is working directly on the pointer.

pkg/sbom/diff_nodelist.go
```diff
@@ -0,0 +1,256 @@
...
+	}
+
+	return &removedNodeList
+}
+
+// NodeListDiffNodes represents the differences between two NodeList nodes.
+type NodeListDiffNodes struct {
+	Added    []*Node
+	Removed  []*Node
+	NodeDiff []*NodeDiff
```
puerco, line 54: This data structure seems problematic to me. As the NodeDiff array gets populated, the original context of the modified node gets lost. In other words, we can know that some nodes had changes but we don't know which.

pkg/sbom/nodelist.go
```diff
@@ -266,6 +266,27 @@ func copyEdgeList(original []*Edge) []*Edge {
 	return nodeCopy
 }
 
+// Copy returns a duplicate of the NodeList.
+func (nl *NodeList) Copy() *NodeList {
+	nlo := &NodeList{
+		// 2DO not sure what to do.
+		// state:         nl.state,
+		// sizeCache:     nl.sizeCache,
+		// unknownFields: nl.unknownFields,
```
puerco, line 275: We've dropped thse from other copy funcs
```suggestion
```

puerco (commented):
Thanks for this huge effort @houdini91! I had some time to give this one a first pass and I left some comments. 

I'm a little bit on the fence about how useful the resulting structure is.  It seems to me that we may want something more like the return values of `AddedNodeList()` and `RemovedNodeList()` but I can almost guarantee that those two will return invalid nodelists. Additionally, the struct has a problem where node changes will get lost (see comment).

I would suggest stashing the qualifier stripping options from this, PR (see notes below), we need a more scalable mechanism to mutate elements when diffing and that could be a whole pr on its own.

What do you think if we try to think about the diff results from a use-case perspective? What I mean is trying to think about cases where a NodeList diff would useful and seeing if the resulting struct fits the purpose. Let's chat more.

pkg/sbom/nodelist.go
houdini91: Resolved by merging in main back to branch.

pkg/sbom/diff_nodelist.go
houdini91: removed strip hack.
It was put in to be able to build a simple version drift command thought it basic enough a functionality to include out of the box as a option but i agree a more generic approach is needed.

pkg/sbom/diff_nodelist.go
houdini91: The equal API to return a integer to be able to iterate the two list.
Maybe we can add a API for compare? 
```
func (e *Edge) Compare(e2 *Edge) int {
	if e2 == nil {
		return 1
	}
	return strings.Compare(e.flatString(), e2.flatString())
}
```

houdini91:
>What do you think if we try to think about the diff results from a use-case perspective? What I mean is trying to think about cases where a NodeList diff would useful and seeing if the resulting struct fits the purpose. Let's chat more.

Agreed the result structure is more based on my imp process than use case.
I did use it for Version drift so for that one it fine.
Another use case we like this to work for is to track losing information in our writers (haven't tried it yet).

In general currently the result sturucture is something like

```
type NodeListDiff struct {
	NodesDiff         NodeListDiffNodes
	EdgesDiff         NodeListDiffEdges
	RootElmementsDiff NodeListRootElementDiff
}
``
Is not a data structure that is connected to the original documents it more a object diff for the nodes, edges and roots.

I was thinking to go the node list way but as you mention it will be a partial node list.
```
type NodeListDiff struct {
	Added       NodeList
	Removed     NodeList
	DiffAdded   NodeList
	DiffRemoved NodeList
}

var diffRes = NodeListDiff{
	// Added partial Node list
	Added: NodeList{
		Nodes:        []*Node{},  // NodeListDiffNodes.Added
		Edges:        []*Edge{},  // NodeListDiffEdges.Added
		RootElements: []string{}, // NodeListRootElementDiff.Added
	},
	// Removed partial Node list
	Removed: NodeList{
		Nodes:        []*Node{},  // NodeListDiffNodes.Removed
		Edges:        []*Edge{},  // NodeListDiffEdges.Removed
		RootElements: []string{}, // NodeListRootElementDiff.Removed
	},
	// Partial node list of same node with modification.
	DiffAdded: NodeList{
		Nodes: []*Node{}, // NodeDiff.Added
	},
	DiffRemoved: NodeList{
		Nodes: []*Node{}, // NodeDiff.removed
	},
}
```

File: pkg/sbom/diff_nodelist.go (hunk 5 above)
Comment (houdini91) [reply pkg/sbom/diff_nodelist.go:54]: Agreed,
Maybe we encapsulate the original Documents in the result as and somehow reference the original node.

I am open to your suggestion though.

File: pkg/sbom/diff_nodelist.go (hunk 4 above)
Comment (houdini91) [reply pkg/sbom/diff_nodelist.go]: removed the hack

File: pkg/sbom/diff_nodelist.go (hunk 3 above)
Comment (puerco) [reply pkg/sbom/diff_nodelist.go]: Cool, we will have methods for this but it is important to scope out the identifier matching logic from the diffin code.

File: pkg/sbom/diff_nodelist.go (hunk 5 above)
Comment (puerco) [reply pkg/sbom/diff_nodelist.go:53]: I think this can be solved by extending the NodeDiff to capture the original node's ID. The structure would end up something like:

```golang 
type NodeDiff struct {
	NodeID string
	Added     *Node
	Removed   *Node
	DiffCount int
}
```
Since we are looking at different scenarios where we may be diffing nodes with different IDs (such as the case you had, diffing nodes with related purls ). For those cases, we may want to ensure we are capturing the ID from nl2 in the `NodeDiff.Added` result.

houdini91: Ok Cool, So we end up just adding a NodeID to the NodeDiff? 
As For retaining the Node ID i think that because the diff operates on the Node.ID as a index it should always be present.

houdini91:
@puerco Do we expect user of the diff object to refer from NodeId in the Diff Object back to the respective documents in order to query the entirety of the information.
Do we want to imp such a API in a higher level? 
Another option can be to retain the equal parts in our Diff Object which should give us our missing infromation ?

pkg/sbom/diff_nodelist.go
```diff
@@ -0,0 +1,234 @@
...
+	})
+
+	sort.Slice(diff.NodeDiff, func(i, j int) bool {
+		return strings.Compare(diff.NodeDiff[i].Added.flatString(), diff.NodeDiff[j].Added.flatString()) <= 0
+	})
+
+	return diff
+}
+
+func (nl *NodeList) diffEdges(nl2 *NodeList) NodeListDiffEdges {
```
puerco, line 179: This function is a nice start but it needs a little bit more work.

To generate the edge diff we need to compare and extract individual graph connections sharing the origin and type. For example, if I diff these two:

```go
Edge{
  "from": "node1"
  "type": "contains"
  "to": []string{"nodeA", "nodeB"}
 } 
Edge{
  "from": "node1"
  "type": "contains"
  "to": []string{"nodeA"}
 }
```
I should get the following diff:
```go
EdgeSetDiff{
  Removed: []Edge{
    Edge {
      "from": "node1"
      "type": "contains"
      "to": []string{"nodeB"}
    },
 }
```

 Also, we may need to work on an edge consolidation function, see this [comment about consolidation](https://github.com/protobom/storage/pull/4#issuecomment-2148705949) on the storage repo.

puerco:
OK, nice work @houdini91 . It's looking good! I added some comments to let us move forward.

I think the data structure (with the `NodeDiff` change suggested above) is enough to hold all changes and make the `NodeListDiff` effective. The key here is to think of it as if this were a regular patch: A patch (diff) file needs enough context so that it can be applied. If a patch is missing line numbers, chunk data, byte offsets, filenames, etc I may not be able to apply it. The data structure needs to be able to hold enough context for it to be reapplied against a NodeList. Tests should expose any limitations in the diff formats.

The next two challenges are the implementation and building super robust testing. The edge diff needs more work, see the comment above. I want to do a deeper pass on the code to check the algorithms but the suggestions above should get this moving, thanks Mikey!

puerco:
> Do we expect user of the diff object to refer from NodeId in the Diff Object back to the respective documents

No, the diffing of nodelists should never need a document (or its metadata section), they need to be independent. In addition to the comment above I think you can summarize the diff as two things:

1. context to find the position where it should be applied
2. The changed data in a format that can be used to write the original.

The diff should only hold the delta, no other information from the original should be required.

houdini91:
@puerco Up

puerco:
We still need this!

houdini91:
@puerco Upstream main.

manifestori:
looking forward to it @houdini91

houdini91:
@puerco

houdini91:
@puerco Gently reminder to review/approve the PR when you get a chance.

pkg/sbom/diff.go
```diff
@@ -7,6 +7,7 @@ import (
 )
 
 type NodeDiff struct {
+	NodeId    string
```
puerco, line 10: Why do we need the NodeId here? I think most of the diff'ing scenarios will probably have different IDs for equivalent nodes, in other words one SBOM describing `packageA` will have an ID in the node while a second SBOM describing the same `packageA` will have a different node ID.

I think we should improve the APIs to find equivalent nodes instead of matching the IDs as we do now. Matching by Ids is easy but most likely will not work on most real world scenarios. 

Looking at the diff code, this ID is not really used. Should we drop it? What do you think?

pkg/sbom/diff_nodelist.go
puerco: OK, sounds good. Do you want to add this in a follow-up ?

houdini91:
@puerco 
>OK, sounds good. Do you want to add this in a follow-up ?

Could you clarify what you mean? 

FYI: I believe we should unify edges into a consistent format—such as when A → B and A → C—before performing the comparison. I’ve noticed this issue can lead to misalignment in the diff results. In our current serializers, we’d need to fix this, unless we plan to keep the dependency format exactly as it appears in the original document.”

PR #174, opened 2024-01-22
GORM model generation

jhoward-lm:
Update `api/sbom.proto` with additional options specific to generating GORM models. New models are written to a separate file, `pkg/sbom/sbom.pb.gorm.go`.


hack/verify-proto.sh
```diff
@@ -10,11 +10,13 @@ source hack/common.sh
...
 export PROTOC_GO_VER="v1.31.0"
+export PROTOC_GORM_VER="v1.1.3-0.20231122062459-d3024d4fa7c9"
 
 curl -LO https://github.com/protocolbuffers/protobuf/releases/download/v${PROTOC_VER}/protoc-${PROTOC_VER}-linux-x86_64.zip
 unzip protoc-${PROTOC_VER}-linux-x86_64.zip -d $HOME/.local
 go install google.golang.org/protobuf/cmd/protoc-gen-go@${PROTOC_GO_VER}
+go install github.com/infobloxopen/protoc-gen-gorm@${PROTOC_GORM_VER}
 
-$HOME/.local/bin/protoc --proto_path=$HOME/.local/include --proto_path=. --go_out=pkg ./api/sbom.proto
+$HOME/.local/bin/protoc --proto_path=$HOME/.local/include --proto_path=$HOME/go/pkg/mod --proto_path=. --go_out=pkg --gorm_out=pkg ./api/sbom.proto
```
puerco, line 20: Why the added `proto_path` path?

pkg/sbom/sbom.pb.go
```diff
@@ -101,11 +102,11 @@ func (x HashAlgorithm) String() string {
 }
 
 func (HashAlgorithm) Descriptor() protoreflect.EnumDescriptor {
-	return file_api_sbom_proto_enumTypes[0].Descriptor()
+	return file_sbom_proto_enumTypes[0].Descriptor()
```
puerco, line 105: The protoc invocation is renaming these functions which is causing the diff that breaks the test. I think what is happening here is that for some reason, the protoc invocation that generated this file ran effectively from the `api/` dir and not from the repo root.

The same test should also cover the newly added gorm types.

hack/verify-proto.sh
jhoward-lm: Reverted

pkg/sbom/sbom.pb.go
jhoward-lm: Fixed the invocation. Still need to rework the gorm options so that the other new content (e.g. the new `Document.Id` field accessor) isn’t written to the original sbom.pb.go. 

Since it’s adding a new file though, will the `git diff` check in the build test pass?

puerco:
Oh no, it looks like the protoc gorm generator output is not deterministic :/
```
-	Contacts          []*PersonORM `gorm:"foreignKey:Name;references:Url;many2many:person_contacts;joinForeignKey:PersonName;joinReferences:ContactUrl"`
+	Contacts          []*PersonORM `gorm:"foreignKey:Phone;references:Phone;many2many:person_contacts;joinForeignKey:PersonPhone;joinReferences:ContactPhone"`
```

jhoward-lm:
> Ah no looking closely they are different fields, the checked in generated code is different than the one generated in the test.

This should be fixed now. I also reverted some things that were unintentionally modified

veramine:
Hi @jhoward-lm thanks for the PR.  

I'm not familiar with gorm but just went to look it up.  This adds dependencies to (amongst others):

github.com/jinzhu/gorm v1.9.16 // indirect
github.com/jinzhu/inflection v1.0.0 // indirect
github.com/jinzhu/now v1.1.5 // indirect

Looks like github.com/jinzhu/gorm hasn't had any updates in several years and the author points to https://github.com/go-gorm/gorm instead, which has been regularly updated.

Sorry for the newb question, but is go-gorm/gorm similar enough to jinzhu/gorm that we should consider using that when adding this new functionality to protobom?  

If there is a reason for using jinzhu/gorm, it's ok with me to use old code because I think this is functionality worth adding.  Thank you for contributing it.  But if we can use go-gorm/gorm instead of the seemingly abandoned jinzhu/gorm, we may want to think about doing that instead.

Thanks again!

jhoward-lm:
@puerco The direct dependency I added was `gorm.io/gorm`. Looking at its [go.mod file](https://github.com/go-gorm/gorm/blob/master/go.mod) (from the `github.com/go-gorm/gorm` project you mentioned), I think that must be what added the indirect dependencies

veramine:
We discussed this issue in our protobom group meeting this morning.  This feature makes sense, it was a good idea.  Thanks for contributing!

pkg/sbom/sbom.pb.go
puerco: (we discussed this on slack: the test passes and it covers the new generated code, thanks!)

jhoward-lm: I didn't push this change, but limiting the `git diff` check in `verify-proto.sh` to just the compiled protobufs made it a little easier to test locally.

```shell
git diff --exit-code -- **/*.pb{,.gorm}.go || exit_with_msg ...
```

jhoward-lm:
Is there anything that still needs to happen to call this done? For the open conversations, do I mark those as resolved or is that for the commenter to do so they can verify?

@puerco we talked on slack about marking this new functionality as experimental, do you want that to go in this PR?

cc @veramine

jhoward-lm:
@puerco @veramine 

Couple of follow-on questions from my previous comment:

- How would you feel about a new directory `pkg/sbom/beta` and put the generated GORM code in there?
- Would you be ok with a symlink inside the new folder (e.g. `pkg/sbom/beta/sbom.pb.go -> ../sbom.pb.go`) so that the file the GORM code imports doesn't need to be tracked twice?

I have all this done locally, but I didn't want to push it without consensus, or if it's going to hold up the PR. The reason for the symlink is because I wasn't able to get `protoc` to write the generated files to two separate locations and update the Go import paths correctly.

puerco:
I'm fine merging it in ./beta and having the symlink. I think the goal would be to graduate it out of beta soon, though.

If this makes it easy to iterate quickly I'm all for it.

jhoward-lm:
Couldn't get the symlink to work correctly, got this error when testing locally:

```console
panic: proto: file "api/sbom.proto" is already registered
        previously from: "github.com/bom-squad/protobom/pkg/sbom"
        currently from:  "github.com/bom-squad/protobom/pkg/sbom/beta"
See https://protobuf.dev/reference/go/faq#namespace-conflict
```

I ended up having to do sort of a hacky workaround. Take a look at the most recent commit when you get a chance and let me know if that will be sufficient or not

veramine:
I'm ok with that kind of creativity in the Makefile, especially for rarely used targets.  make proto is not going to happen very often.  However, in this specific case, on my first try testing I got this:

```
% make
conformance-test         Run the conformance test suite
conformance              Regenerate the conformance golden samples. Only to be run when changes are expected
fakes                    Rebuild the fake implementations
proto                    Rebuild protobuf autogenerated code
% make proto 
protoc --go_out=pkg api/sbom.proto
gorm.proto: File not found.
api/sbom.proto:3:1: Import "gorm.proto" was not found or had errors.
make: *** [proto] Error 1
```

jhoward-lm:
@veramine Yeah, I left it out because the other dependencies being installed in `verify-proto.sh` (like the versioned `protoc` and the `go install` commands) weren't previously included in the `Makefile` either. Do you want me to add it?

veramine:
Aha, I just had to run verify-proto first.  Got it.  Okay now I got further:

```
% find . -name gorm.proto
./gorm.proto
% make proto
protoc --go_out=pkg api/sbom.proto
mkdir -p beta
cp api/sbom.proto beta
sed -i -E 's/^(package bomsquad\.protobom)/\1\.beta/' beta/sbom.proto
sed: 1: "s/^(package bomsquad\.p ...": \1 not defined in the RE
make: *** [proto] Error 1
```

let me see if I can figure this one out...

jhoward-lm:
Are you on MacOS? It might be a GNU vs POSIX issue with the `sed` command.  I had installed `gnu-sed` with Homebrew. I should probably rewrite it to be more cross-compatible.

veramine:
Yep, I'm on a mac.  My version of sed could be super duper old.  A simpler version of the replacement worked but I can appreciate you sniping super specifically to make sure there is nothing else changed..  

```
% sed -i -E 's/package bomsquad\.protobom/package bomsquad\.protobom.beta/' beta/sbom.proto
% head -7 beta/sbom.proto
syntax = "proto3";

import "gorm.proto";
import "google/protobuf/timestamp.proto";

option go_package = "sbom/";
package bomsquad.protobom.beta;
```

jhoward-lm:
@veramine I updated the sed command to write directly to the copied file location instead of copying it then trying to modify it in-place. 

```shell
sed -E 's/^(package bomsquad\.protobom)/\1\.beta/' api/sbom.proto > beta/sbom.proto
```

Did you want me to not use the RE back-reference (`\1`)?

veramine:
oh goodness, I have no preference at all on the regexp.  Your new sed line worked for me, thanks!

houdini91 (approved):
LGTM, thanks for the awesome PR,
My only suggestion is add some short of documentation for using the new beta API.
I would suggest godoc example or some other static readme.

jhoward-lm:
@houdini91 I added a [godoc example](https://github.com/bom-squad/protobom/pull/174/commits/aa24acee465fee3c919da5c8307c398b7122b00d#diff-6d5c543c972dc5579d5a83de6f3bbf284869e85e385ebb5b260894c40de49887) if you want to check it out. Also, I had to scrap the `beta` folder thing. Since the `pkg/sbom/beta/sbom.pb.go` was a separate copy of the one in the parent folder, the new GORM models only work with the types in the beta package, and helper functions like `reader.ParseFile()` were returning the incompatible `Document` type from the parent folder. I was also able to make some minor improvements like `uint32` primary keys instead of strings.

cc @puerco @veramine

Makefile
```diff
@@ -14,6 +14,19 @@ conformance-test: ## Run the conformance test suite
...
+proto-gorm: gorm.proto proto ## Compile protobufs with protoc-gen-gorm plugin
+	protoc --gorm_out=pkg api/sbom.proto
+
+PROTOC_GORM_VER := d3024d4fa7c9ce9d62e2f2cc0b42faf679211846
+PROTOC_GORM_URL := https://raw.githubusercontent.com/infobloxopen/protoc-gen-gorm/${PROTOC_GORM_VER}/proto/options/gorm.proto
+
+.SILENT: gorm.proto
+gorm.proto:
+	echo Downloading protoc-gen-gorm...
+	go install github.com/infobloxopen/protoc-gen-gorm@$(PROTOC_GORM_VER)
```
fkautz, line 27: Will this override an existing protoc-gen-gorm install? It might be good to make this an explicit command rather than an implicit global side effect.

Makefile
```diff
@@ -14,6 +14,19 @@ conformance-test: ## Run the conformance test suite
...
 
+.PHONY: proto-gorm
+proto-gorm: gorm.proto proto ## Compile protobufs with protoc-gen-gorm plugin
+	protoc --gorm_out=pkg api/sbom.proto
+
+PROTOC_GORM_VER := d3024d4fa7c9ce9d62e2f2cc0b42faf679211846
+PROTOC_GORM_URL := https://raw.githubusercontent.com/infobloxopen/protoc-gen-gorm/${PROTOC_GORM_VER}/proto/options/gorm.proto
+
+.SILENT: gorm.proto
+gorm.proto:
```
fkautz, line 25: Does gorm provide precompiled libraries that we can use instead of importing and doing the code generation ourselves?

jhoward-lm: They do, located [here](https://github.com/infobloxopen/protoc-gen-gorm/blob/d3024d4fa7c9ce9d62e2f2cc0b42faf679211846/options/gorm.pb.go), but I wasn't able to figure out how to access the needed message options (e.g. `option (gorm.opts) = ...`) without importing the proto file.

The proto file gets installed along with the plugin, but I couldn't come up with an intuitive way to find the file path. e.g. with a `PROTOC_GORM_VER` of `d3024d4fa7c9ce9d62e2f2cc0b42faf679211846`, on my system the file ends up located at `~/go/pkg/mod/github.com/infobloxopen/protoc-gen-gorm@v1.1.3-0.20231122062459-d3024d4fa7c9` after the `go install`.

Makefile
jhoward-lm: It installs it into `$GOPATH/pkg/mod`, but doesn't update go.mod for the project. Multiple versions of protoc-gen-gorm can be installed simultaneously so it won't be overridden

Makefile
jhoward-lm: @fkautz Do you know of a way to use the protoc-gen-gorm options without importing the proto? How do you want me to resolve this conversation?

jhoward-lm:
@puerco @fkautz @veramine @houdini91

Any updates on this, or is there anything I still need to do?

houdini91 (approved):
Lgtm

Makefile
houdini91: @veramine @jhoward-lm Here's my understanding:

Gorm lacks a straightforward method for importing the gorm.proto. They suggest incorporating it into a build process, as you've done. Previously, with older Go versions, the package could be fetched into GOPATH/pkg/src, allowing you to map it during protoc builds.

My suggestion is to use a .tmp directory for build files that should be ignored. Additionally, I prefer having the entire repository cloned, as it contains many other protos we may need to reference.

I've attached an example patch (although I lack permission to suggest it to this branch). What are your thoughts?

[gorm_import.patch.txt](https://github.com/bom-squad/protobom/files/14330359/gorm_import.patch.txt)

Makefile
houdini91: While it's not obligatory for this pull request, I propose we move it to .tmp (refer to my patch in an alternative format) with a prefix:```
GOBIN=$(shell realpath $(TEMPDIR)) go install ..
```
This approach directs all build tools to the designated temporary folder instead of relying on the local GOPATH. Although it might not make a difference in the development container, I believe it's worth considering.

@puerco, what are your thoughts on installing all our tools under .tmp within the local repository?

Comment (houdini91) [reply Makefile:27]: @manifestori

File: Makefile (hunk 4 above)
Comment (jhoward-lm) [reply Makefile:25]: Would something like this do the same thing as cloning with `git`? I’m not exactly a Go expert, maybe `GOBIN` instead of `GOPATH`?

```makefile
ROOT_DIR ?= $(abspath $(dir $(lastword $(MAKEFILE_LIST))))

GOPATH=$(ROOT_DIR)/.tmp go install protoc-gen-gorm@$(PROTOC_GORM_VER)
```
AFAIK, the `protoc-gen-gorm` project really only has 2 `.proto` files meant to be imported by other protobufs ([link](https://github.com/infobloxopen/protoc-gen-gorm/tree/main/proto)). Since this PR adds it as a project dependency, I would think that all of the files in that repo should be locally available to devs once `go install` is run.

Here’s the patch text in a ` ```diff` code block:

<details>

<summary>Expand</summary>

```diff
diff --git a/.gitignore b/.gitignore
index 20c3236..ab39967 100644
--- a/.gitignore
+++ b/.gitignore
@@ -20,3 +20,4 @@
 # Go workspace file
 go.work
 .vscode
+*.tmp
diff --git a/Makefile b/Makefile
index f9d5ac0..2039df2 100644
--- a/Makefile
+++ b/Makefile
@@ -1,3 +1,13 @@
+
+TEMPDIR = ./.tmp
+
+ifndef TEMPDIR
+	$(error TEMPDIR is not set)
+endif
+
+$(TEMPDIR):
+	mkdir -p $(TEMPDIR)
+
 .PHONY: help
 help:
 	@grep -Eh '^[a-zA-Z_-]+:.*?## .*$$' $(MAKEFILE_LIST) | sort | awk 'BEGIN {FS = ":.*?## "}; {printf "$(BOLD)$(CYAN)%-25s$(RESET)%s\n", $$1, $$2}'
@@ -14,18 +24,21 @@ conformance-test: ## Run the conformance test suite
 proto: ## Rebuild protobuf autogenerated code
 	protoc --go_out=pkg api/sbom.proto
 
-.PHONY: proto-gorm
-proto-gorm: gorm.proto proto ## Compile protobufs with protoc-gen-gorm plugin
-	protoc --gorm_out=pkg api/sbom.proto
+.PHONY: gorm proto
+proto-gorm: gorm ## Compile protobufs with protoc-gen-gorm plugin
+	protoc --go_out=pkg --gorm_out=. api/sbom.proto -I. -I$(TEMPDIR)
 
 PROTOC_GORM_VER := d3024d4fa7c9ce9d62e2f2cc0b42faf679211846
 PROTOC_GORM_URL := https://raw.githubusercontent.com/infobloxopen/protoc-gen-gorm/${PROTOC_GORM_VER}/proto/options/gorm.proto
 
-.SILENT: gorm.proto
-gorm.proto:
+.SILENT: gorm
+gorm:
 	echo Downloading protoc-gen-gorm...
+	rm -rf $(TEMPDIR)/protoc-gen-gorm | true
+	git clone https://github.com/infobloxopen/protoc-gen-gorm.git $(TEMPDIR)/protoc-gen-gorm --depth 1 --branch main --single-branch
 	go install github.com/infobloxopen/protoc-gen-gorm@$(PROTOC_GORM_VER)
-	curl --fail --silent --show-error --location --remote-name --url $(PROTOC_GORM_URL)
+	
+# curl --fail --silent --show-error --location --remote-name --url $(PROTOC_GORM_URL)
 
 .PHONY: fakes
 fakes: ## Rebuild the fake implementations
diff --git a/api/sbom.proto b/api/sbom.proto
index 24b3b07..896c5c9 100644
--- a/api/sbom.proto
+++ b/api/sbom.proto
@@ -1,7 +1,8 @@
 syntax = "proto3";
 
-import "gorm.proto";
+
 import "google/protobuf/timestamp.proto";
+import "protoc-gen-gorm/proto/options/gorm.proto";
 
 option go_package = "sbom/";
 package bomsquad.protobom;
```

</details>

I would probably do `$(RM) -r` and have it in a separate clean target but it’s just a style thing

houdini91: If it functions correctly, then yes, I attempted it with the latest version of Go but couldn't get the sources to be fetched by the Go binary, so I resorted to pulling them with Git.

As far as I understand, GOPATH is a somewhat backward-compatible feature, whereas GOBIN is typically just `$GOPATH/bin`. So, in reality, I don't think it will make much of a difference.

Regarding the Git clone, I suppose we can only fetch the protocol, but I thought it would be clearer how to add other protocol source repositories if needed when cloning the entire repository.

Yes, indeed, we can remove the `rm` command and instead move it to a different make target. I just outlined it quickly.

jhoward-lm:
@puerco @veramine @houdini91  @fkautz

looks like the golangci-lint failures are from a file I didn't touch, serializer_cdx.go. How do you want me to address it?

houdini91:
Yeah I noticed this in all my prs

veramine:
I just made the fix for the lint issue in your branch, hope you don't mind!

jhoward-lm:
> I just made the fix for the lint issue in your branch, hope you don't mind!

Not a problem at all, thanks!

puerco (approved):
All right thanks a lot @jhoward-lm for this awesome contribution! Sorry for taking so long to approve, things should start moving at their regular pace now :rocket: 

/lgtm

PR #193, opened 2024-03-26
Generate ent schemas and code

jhoward-lm:
## Description

This PR is to provide an example of what an `ent` representation of protobom might look like.

It looks like a lot of changes, but these files were the only ones modified or added manually:

### Modified

- .golangci.yml
- Makefile
- pkg/formats/sniffer.go (fix an unrelated lint finding)

### New

- ent/Makefile
- ent/entc.go
- ent/generate.go
- ent/schema/*
- hack/verify-ent.sh
- pkg/native/unserializers/beta/unserializer_ent_cdx.go
- pkg/native/unserializers/beta/unserializer_ent_cdx_test.go
- pkg/native/unserializer_ent.go

## Summary

- Duplicated `sbom.proto` in `ent/proto/entpb/` with `protoc-gen-ent` specific message and field options
  - `protoc-gen-ent` is unable to generate schemas for the current form of `sbom.proto` due to lack of support for the `optional` keyword and map field types
- Schemas generated from protobuf definitions into `ent/schema`
- Generate ent code from the schemas in `ent/generate.go`

PR #194, opened 2024-03-27
Remove GORM generated code in preparation for ent

jhoward-lm:
From discussion in protobom project meeting, separate PR to remove generation of GORM code from project prior to [ent integration PR](https://github.com/bom-squad/protobom/pull/193).


puerco (approved):
Thanks !!

PR #204, opened 2024-04-18
refactor to use sync.Map and sync.Once for writer

naveensrinivasan:
- refactored the code to use sync.Once and sync.Map primitives

### DIFF
--- a/pkg/writer/writer.go
+++ b/pkg/writer/writer.go
@@ -18,8 +18,8 @@ type Writer struct {
 }
 
 var (
-	regMtx         sync.RWMutex
-	serializers    = make(map[formats.Format]native.Serializer)
+	serializers    sync.Map
+	once           sync.Once
 	defaultOptions = &Options{
 		RenderOptions: &native.RenderOptions{
 			Indent: 4,
@@ -30,6 +30,7 @@ var (
 )
 
 func New(opts ...WriterOption) *Writer {
+	ensureSerializersInitialized()
 	w := &Writer{
 		Options: defaultOptions,
 	}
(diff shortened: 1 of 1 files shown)


pkg/writer/writer.go
```diff
@@ -30,6 +30,7 @@ var (
 )
 
 func New(opts ...WriterOption) *Writer {
+	ensureSerializersInitialized()
```
puerco, line 33: Why move the load function here instead of doing it on init() ? It needs to be called in lots of places and makes tracking it difficult. Also, if anything creates a writer without using New() it would not have access to the serializers.

puerco (commented):
Thanks for this @naveensrinivasan , one question 👇

pkg/writer/writer.go
naveensrinivasan: @puerco  The reason for moving away from the `init` func is they are hard to test, and it makes it harder for others to keep track of the initialization. I would recommend no code should create without using `New()`

https://github.com/leighmcculloch/gochecknoinits?tab=readme-ov-file

Thoughts? 

Do you have strong opinion on keeping the init?

naveensrinivasan: @puerco  a friendly ping

puerco: Sorry @naveensrinivasan I forgot to follow up on this one.

I think we're fine switching to your proposed initialization mechanics. Let me explain first what I thought when I read it:

My main concern here was that by using `init()` the initialization of the default plugins is automatic, it always happens. Moving the plugin load logic to `ensureSerializersInitialized()` means we must remember to load them when needed. This means unexpected behavior when you forget to load the serializers/unserializers

Second, the main argument about not using `init()` is to avoid a global state. This adds the complexity to initialise the loaded serializers manually but still keeps the global state (the `serializers` map).

_**However**_, after giving it some thought...

Since the `serializers` map is local to the package, all code paths that need to read (or set) the serializers collection _must_ go through methods and functions in the package where we have full control. In other words, all instances of failures trying to access an uninitialized `serializers` collection will be bugs in the writer package due an oversight on our part (and not the application importing protobom).

So, I say let's ship it :rocket:

puerco (approved):

PR #208, opened 2024-04-22
Fix panic when writing to a nonexistent file

eddiezane:
This PR fixes a panic that would occue when attempting to write to a nonexistent file.


puerco (approved):
Asesome thanks @eddiezane!

PR #210, opened 2024-04-23
Storage interface calls + simple FS backend

puerco:
This is a work in progress pull request that builds storage support into protobom as defined in the Protobom Persistence System document: https://docs.google.com/document/d/1Us0EMnw0bT6w24CZKHPJi1_BgQZ77zbfbSc55qymUQ0/edit#heading=h.v5t0rukhe4kx

With this change, we are introducing the following changes:

- [x] Added `write.Store()` function
- [x] Added storage options to writer options
- [x] Added `reader.Retrieve()` function
- [x] Added storage options to reader options
- [x] Simple default filesystem storage backend
- [x] filesystem storage backend tests

~Depends on https://github.com/protobom/storage/pull/2~

This PR will be updated as the storage backend module and design document are updated.

Signed-off-by: Adolfo García Veytia (Puerco) <puerco@stacklok.com>


pkg/reader/reader.go
```diff
@@ -150,3 +154,28 @@ func (r *Reader) detectFormat(rs io.ReadSeeker) (formats.Format, error) {
 	}
 	return format, nil
 }
+
+// Retrieve reads a document from the configured storage backend using the
+// default options.
+func (r *Reader) Retrieve(id string, o *Options) (*sbom.Document, error) {
+	return r.RetrieveWithOptions(id, defaultOptions)
```
mikhailswift, line 161: should this use `o`, and fall back to `defaultOptions` if `o` is nil?

mikhailswift: Or, I suppose after reading the comment, should `o` be dropped as a parameter

mikhailswift (approved):
just one small comment, but nothing major. this looks good!

pkg/reader/reader.go
puerco: Oooh nice catch, thank you @mikhailswift!!! :

PR #211, opened 2024-04-23
Fix missing relationships in nodelist.RelateNodeAtID()

puerco:
This commit fixes a bug that was causing edges from the source NodeList
to be lost when calling `nodelist.RelateNodeAtID()`.

Signed-off-by: Adolfo García Veytia (Puerco) <puerco@stacklok.com>

### DIFF
--- a/pkg/sbom/edge.go
+++ b/pkg/sbom/edge.go
@@ -248,3 +248,21 @@ func (e *Edge) flatString() string {
 	sort.Strings(tos)
 	return e.From + ":" + e.Type.String() + ":" + strings.Join(tos, "+")
 }
+
+// AddDestinationById adds identifiers to the destination list of the edge. The
+// new destination identifiers are guaranteed to be added only once and will
+// not be duplicated if there is already a destination with the same ID.
+func (e *Edge) AddDestinationById(ids ...string) {
+	dests := map[string]struct{}{}
+	for _, id := range e.To {
+		dests[id] = struct{}{}
+	}
+
+	for _, id := range ids {
+		if _, ok := dests[id]; ok {
+			continue
+		}
+		dests[id] = struct{}{}
+		e.To = append(e.To, id)
+	}
+}
--- a/pkg/sbom/nodelist.go
+++ b/pkg/sbom/nodelist.go
@@ -666,7 +666,7 @@ func (nl *NodeList) RelateNodeListAtID(nl2 *NodeList, nodeID string, edgeType Ed
 		nl.Edges = append(nl.Edges, edge)
 	} else {
 		// Perhaps we should filter these
-		edge.To = append(edge.To, nl2.RootElements...)
+		edge.AddDestinationById(nl2.RootElements...)
 	}
 
 	for _, n := range nl2.Nodes {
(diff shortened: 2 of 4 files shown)


pkg/sbom/nodelist.go
```diff
@@ -676,6 +676,21 @@ func (nl *NodeList) RelateNodeListAtID(nl2 *NodeList, nodeID string, edgeType Ed
 		nl.AddNode(n)
 	}
 
+	// Copy the remaining edges from n2
+	for _, e := range nl2.Edges {
+		// Check if we have an edge of the samer type already in the
+		// nodelist and if so, reuse it:
+		if _, ok := nlEdges[e.From]; ok {
+			if _, ok2 := nlEdges[e.From][e.Type]; ok2 {
+				nlEdges[e.From][e.Type][0].AddDestinationById(e.To...)
```
eddiezane, line 685: Can you please explain why you are grabbing the first one out of this From.Type collection?

pkg/sbom/edge.go
```diff
@@ -248,3 +248,21 @@ func (e *Edge) flatString() string {
...
 	return e.From + ":" + e.Type.String() + ":" + strings.Join(tos, "+")
 }
+
+// AddDestinationById adds identifiers to the destination list of the edge. The
+// new destination identifiers are guaranteed to be added only once and will
+// not be duplicated if there is already a destination with the same ID.
+func (e *Edge) AddDestinationById(ids ...string) {
+	dests := map[string]struct{}{}
+	for _, id := range e.To {
+		dests[id] = struct{}{}
```
eddiezane, line 258: Will this ever blow away existing things?

pkg/sbom/edge.go
```diff
@@ -248,3 +248,21 @@ func (e *Edge) flatString() string {
...
+	}
+
+	for _, id := range ids {
+		if _, ok := dests[id]; ok {
+			continue
+		}
+		dests[id] = struct{}{}
+		e.To = append(e.To, id)
```
eddiezane, line 266: I am not understanding this

eddiezane (commented):
I'd like to see a test that fails @ main but passes with this change.

puerco:
I've noticed that `nodelist.RelateNodeAtID()` had no test, I added one in https://github.com/protobom/protobom/pull/211/commits/db50bd113421d616ef7bfe651a9bc020bb2c1d84 that should catch this and ensure we're merging properly. We can add more test cases.

pkg/sbom/nodelist.go
puerco: In general, edges created by protobom should only have a single edge pairing from + type. Sometimes you'll have more if an SBOM has them defined.

This is taking the destination nodes from the SBOM2 edge and migrating them to the equivalent edge in SBOM1.

If there is more than one edge of the same type in SBOM1 (say from nodeX + CONTAINS), we don't care to which one we add the destinations from node2 as they are all the same logically, we even have a `cleanEdges()` function in the nodelist that collapses the duplicate edges when we need to.

pkg/sbom/edge.go
puerco: No it's just indexing the existing ids in the edge to iterate just once over them.

pkg/sbom/edge.go
puerco: These two lines add the `id` to the index of the known ids (to avoid duplication) and append the `id` to the edges destination.

pkg/sbom/edge.go
eddiezane: Derp I see that now :facepalm:

eddiezane (approved):
LGTM ty!

fkautz (approved):

PR #229, opened 2024-05-22
Fix protobom document identifier in SPDX unserializer

puerco:
This PR fixes the protobom identifier generated after parsing an SPDX document. It used to be fixed to the SPDX identifier (`DOCUMENT`) which is not correct. This PR changes it so that we return the full URI based on the document namespace.

For invalid documents lacking a namespace, we introduce a fallback to generate a namespace under the SPDX namespace.


manifestori (approved):
LGTM

PR #231, opened 2024-05-28
Revert StoreRetriever to only accept Document

jhoward-lm:
In addition to the current protobom types in the type constraint, this PR allows for implementing `Storer` and `Retriever` for **_slices_** of certain of those types. For example:

```go
// Store implements the Storer interface for *sbom.Node.
func (backend *NodeBackend) Store(n *sbom.Node, opts *storage.StoreOptions) error {
	//
}

// Retrieve implements the Retriever interface for *sbom.Node.
func (backend *NodeBackend) Retrieve(id string, opts *storage.RetrieveOptions) (*sbom.Node, error) {
	//
}
```

as well as

```go
// Store implements the Storer interface for []*sbom.Node.
func (backend *NodesBackend) Store(n []*sbom.Node, opts *storage.StoreOptions) error {
	//
}

// Retrieve implements the Retriever interface for []*sbom.Node.
func (backend *NodesBackend) Retrieve(id string, opts *storage.RetrieveOptions) ([]*sbom.Node, error) {
	//
}
```

This is primarily for use with `Retrieve` so that given an ID, the method can return a slice of all objects associated with that ID, but with an added bonus for `Store`: with the ability to accept a slice of objects, `Store` will be able to insert all ent objects into the database in one bulk transaction, which should be a significant performance improvement over inserting them one at a time.

### DIFF
--- a/pkg/reader/options.go
+++ b/pkg/reader/options.go
@@ -5,7 +5,6 @@ import (
 
 	"github.com/protobom/protobom/pkg/formats"
 	"github.com/protobom/protobom/pkg/native"
-	"github.com/protobom/protobom/pkg/sbom"
 	"github.com/protobom/protobom/pkg/storage"
 )
 
@@ -70,7 +69,7 @@ func WithSniffer(s Sniffer) ReaderOption {
 	}
 }
 
-func WithStoreRetriever(sb storage.StoreRetriever[*sbom.Document]) ReaderOption {
+func WithStoreRetriever(sb storage.StoreRetriever) ReaderOption {
 	return func(r *Reader) {
 		if sb != nil {
 			r.Storage = sb
--- a/pkg/reader/reader.go
+++ b/pkg/reader/reader.go
@@ -60,7 +60,7 @@ func GetFormatUnserializer(format formats.Format) (native.Unserializer, error) {
 
 type Reader struct {
 	sniffer Sniffer
-	Storage storage.StoreRetriever[*sbom.Document]
+	Storage storage.StoreRetriever
 	Options *Options
 }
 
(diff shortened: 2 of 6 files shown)


jhoward-lm:
@puerco @eddiezane @houdini91 requesting review

houdini91 (approved):

puerco:
Mmh what's the use case here? By definition, an id string must identify a single document or protobom node. In fact, for this to work we probably will need to have a schema to overload the ID string with a document namespace:

```golang
x.Retrieve(id string, opts *storage.RetrieveOptions) ([]*sbom.Node, error) 
```

All node IDs are unique to the document namespace. There should be no way to have more than two results from looking up a document from its ID string (it's the promise of the storage backend).

jhoward-lm:
@puerco I can get rid of the type aliases and just put all of the types and slice types in the type constraint directly if that makes it more readable. Meaning:

```go
type (
	ProtobomType interface {
		*sbom.Document | *sbom.Metadata | *sbom.NodeList |
			*sbom.DocumentType | []*sbom.DocumentType |
			*sbom.Edge | []*sbom.Edge |
			*sbom.ExternalReference | []*sbom.ExternalReference |
			*sbom.Node | []*sbom.Node |
			*sbom.Person | []*sbom.Person |
			*sbom.Purpose | []*sbom.Purpose |
			*sbom.Tool | []*sbom.Tool |
			map[sbom.HashAlgorithm]string |
			map[sbom.SoftwareIdentifierType]string
	}

	// ...
)
```

puerco:
@jhoward-lm @eddiezane and me had a deep deep discussion about the original change in this PR and reached the following concussions:

* The current storage interface signatures make the mulit-element `Retrieve()` methods unviable. 
* We don't have any use case to expose a multi-element `Store()` function

As we're rolling back the type constraints on the Retriever interface, we chose to keep feature parity on `Store()` and `Retrieve()` so we're rolling back the whole generic type constraints. It's a micro-breaking change but it was shortlived enough that we expect little impact.

puerco (approved):

PR #247, opened 2024-09-10
fix: README.md to replace parser with unserializer

abhisek:
The `README.md` has broken reference to `docs/parsers.md`. Looking at the code I figured `Unserializer` is what is used by `readers` to read a specific SBOM format into neutral `protobom` object.


cpanato (approved):
lgtm

thanks

puerco (approved):
Thanks!

PR #248, opened 2024-09-15
feat: Adding CycloneDX 1.6 support

idunbarh:
This PR is to update to support CycloneDX 1.6.

The [CycloneDX 1.6 Changelog](https://github.com/CycloneDX/specification/releases/tag/1.6) doesn't show any changes that would impact the protobom implementation resulting in minor changes.
 
This PR includes:

- Updated comments and docs to include CDX 1.6
- Updated implementation to support new version
- Replaced CDX 1.5 with CDX 1.6 in unittests
- Created CDX 1.6 json conformance test


jhoward-lm:
Do we know if there are any data types or fields introduced with 1.6 that didn't previously have an SPDX equivalent but might now?

What I mean is, should the protobom spec be extended with additional types/fields as a result of including the 1.6 spec?

If so, I think that would mean updating all of the serializers/unserializers to account for whatever is added

manifestori (dismissed):
LGTM

api/sbom.proto
```diff
@@ -317,30 +317,30 @@ enum SoftwareIdentifierType {
 // It categorizes the roles that software components can fulfill.
 enum Purpose {
     UNKNOWN_PURPOSE = 0; // Unknown purpose.
-    APPLICATION = 1; // Application purpose. (SPDX2.3, CDX1.5, SPDX3.0)
+    APPLICATION = 1; // Application purpose. (SPDX2.3, CDX1.6, SPDX3.0)
```
puerco, line 320: Thanks @idunbarh this is wonderful. It LGTM, I just have one small nit. Can we drop these bumps in the comments here?

```suggestion
    APPLICATION = 1; // Application purpose. (SPDX2.3, CDX1.5, SPDX3.0)
```

Leaving these in the original version where we first matched them allows us to track compatibility by understanding when we first mapped the property and avoids the diff in the go library.

jhoward-lm (approved):

puerco (approved):
Thank you, lgtm!

PR #263, opened 2024-10-13
Support Properties in the Node

puerco:
This PR adds support for properties in the node. The properties model mimics the CycloneDX model but the field containing the value is called `Data` instead of `Value` because of a name clash with the value scanner generated code.

### Serialization and Unserialization

The CycloneDX serializer and unserializers have been updated to capture and render the properties when importing and exporting to CDX. SPDX is not included for now as it does not support properties natively but will be supported via an upcoming hack.

### Conformance Data

This PR modifies the protos and the Node model. It is not a breaking change but it does augument it resulting in a diff in the conformance test data which is being regenerated.

### Graph API

This PR includes the required changes to fully support properties in the graph API (agumenting, comparing, diff'ing, etc). The functions have been updated as well as their tests.

/cc @jhoward-lm


api/sbom.proto
```diff
@@ -346,3 +349,9 @@ enum Purpose {
     SPECIFICATION = 27; // Specification purpose. (SPDX3.0)
     TEST = 28; // Test purpose. (SPDX3.0)
 }
+
+// Property type
+message Property {
+    string name = 1;    // Name of the property
+    string data = 2;    // Property value
+}
```
jhoward-lm, line 357: Do you still get the name collision if you nest this message inside the `Node` message and set `data` back to `value`?

puerco: Yeah, nesting it only makes the `Property` struct in go to be named  `Node_Property` but its `Value` property will still clash with the `Value()` func from the value scanner.

pkg/native/serializers/serializer_cdx.go
```diff
@@ -404,6 +404,17 @@ func (s *CDX) nodeToComponent(n *sbom.Node) *cdx.Component {
...
+	if len(n.Properties) > 0 {
+		properties := []cdx.Property{}
+		for _, p := range n.Properties {
+			properties = append(properties, cdx.Property{
+				Name:  p.Name,
+				Value: p.Data,
+			})
+		}
+		c.Properties = &properties
+	}
```
jhoward-lm, line 416: I don't fully understand the semantics of a pointer to a slice of values the upstream codebase likes to use, but would this work?

It should skip the for loop if `n.Properties` is empty, so I **_think_** the `if len(n.Properties) > 0` check could be removed. If it gets skipped, it just sets `c.Properties` to the address of the unallocated slice.

```suggestion
	var properties []cdx.Property
	for _, p := range n.Properties {
		properties = append(properties, cdx.Property{
			Name:  p.Name,
			Value: p.Data,
		})
	}

	c.Properties = &properties
```

jhoward-lm: Here's a go playground [link](https://go.dev/play/p/Qzu8LOiJTFi). Hopefully it demonstrates properly what I'm asking about. 

Seems like a prevalent pattern in the serializer/unserializer files, maybe something like this could be applied to all instances of this pattern to improve cognitive complexity scores (and possibly performance)?

puerco: Ah you are right, it's just an old habit I have. I've changed it and repushed.

That `sliceutil` package looks really useful, like something you should release on its own (wink wink)

jhoward-lm:
How would you feel about using the SPDX concept of [annotations](https://github.com/spdx/spdx-spec/blob/ed8c9b520963b651fce3c86422b387e92e6d9a8f/schemas/spdx-schema.json#L237) (also defined at the document level) to also describe CDX properties? It would also "legitimize" the annotations schema that was created in the ent backend.

Thinking `annotator` could map to `Name`, and `comment` to `Value` or something like that

puerco:
@jhoward-lm now that we have the [protobom mod infrastructure](https://github.com/protobom/protobom/pull/264) I will work on exactly that after this one merges :)

jhoward-lm:
> @jhoward-lm now that we have the [protobom mod infrastructure](https://github.com/protobom/protobom/pull/264) I will work on exactly that after this one merges :)

Would that require renaming the `Property` message to `Annotation`? Should it just be named that way to start with?

puerco:
Ah so according to our last discussion, we will not build support for SPDX annotations yet, we will officially support properties (as they are also defined in SPDX3) and write them to annotations in SPDX 2.x if the mod is enabled. But, at least for now, protobom will not have annotation fields.

api/sbom.proto
```diff
@@ -495,6 +498,14 @@ message Person {
   // repeated PersonIdentifier identifiers // Support?
 }
 
+// Property type, key-value store of data.
+message Property {
+  // Name of the property
+  string name = 1;
+  // Property value
```
jhoward-lm, line 505: nit

```suggestion
  string name = 1;

  // Property value
```

jhoward-lm (approved):
Super tiny nit but LGTM

api/sbom.proto
puerco: OK, I've repushed fixing the NL

jhoward-lm (approved):

PR #270, opened 2024-10-16
fix: set sbom reader options.Format to be available in downstream code

ashearin:
Small change to capture document format in sbomReader options to be consumable later. 

example: (current behavior)
```
sbomReader := reader.New()

document, err := sbomReader.ParseStream(bytes.NewReader(sbomData))
if err != nil {
return nil, fmt.Errorf("parsing SBOM data: %w", err)
}

fmt.Println(sbomReader.Options.Format.Type())
```

The Println statement at the end will always return an empty string if the format was not set in the reader options. 

Updated Behavior:
- If o.Format is set, then it will be used with `GetFormatUnserializer` 
- if o.Format is NOT set, we call `detectFormat` and set o.Format to the result and continue on to `GetFormatUnserializer` 

Side effect: This change would allow format information to be accessed later. In the above example, the println would not be empty.

### DIFF
--- a/pkg/reader/reader.go
+++ b/pkg/reader/reader.go
@@ -118,16 +118,15 @@ func (r *Reader) ParseStreamWithOptions(f io.ReadSeeker, o *Options) (*sbom.Docu
 		return nil, fmt.Errorf("options cannot be nil")
 	}
 
-	format := o.Format
 	if o.Format == "" {
 		f, err := r.detectFormat(f)
 		if err != nil {
 			return nil, fmt.Errorf("detecting SBOM format: %w", err)
 		}
-		format = f
+		o.Format = f
 	}
 
-	unserializer, err := GetFormatUnserializer(format)
+	unserializer, err := GetFormatUnserializer(o.Format)
 	if err != nil {
 		return nil, fmt.Errorf("getting format parser: %w", err)
 	}


ashearin:
Still working on updating tests, I have few unit tests failing locally which is unexpected.

puerco (changes requested):
Hey Allen, thanks for working on this! There is a small problem with the current approach. In general, when using the reader, apps should only rely on the autodetect capabilities. You can set the format option but that forces protobom to bypass the autodetection and it makes the reader understand only the format you set. This is intended to be used when autodetection does not work on a particular SBOM and you need to force the reader to use a specific serializer,, for example.

With this change, when you ingest an SBOM, you're altering the options state by setting the `Format` option. This effectively means that once you read an SBOM, the reader will only understand that format from then on. 

When you `fmt.Println(sbomReader.Options.Format.Type())` it returns an empty string meaning autodetect is in place. 

Can you shed some light on your use case to better understand how to surface the data you need?

puerco:
BTW, if you need the format of the original document, it would be a great excuse to prioritize the [source document traceability feature](https://github.com/protobom/protobom/issues/213) we've been discussing.

ashearin:
> Can you shed some light on your use case to better understand how to surface the data you need?

I think you guessed it based on your second comment, but we want to ability to extract the original document format from the sbomReader when adding to the cache in bomctl. 

> BTW, if you need the format of the original document, it would be a great excuse to prioritize the [source document traceability feature](https://github.com/protobom/protobom/issues/213) we've been discussing.

I'll take a look at this feature next week probably, but for now I'll close this PR since it's not the right way to go about getting what I need. The fact that it broke so many tests made me question if my change impacted more than I anticipated so I appreciate the insight!

puerco:
No worries, let me see if I can get some time to get on it! It would be at least good to get the proposal for the new fields going

PR #279, opened 2024-10-31
Refactor source reader

puerco:
With #278 merged, we can now pipe the I/O of the reader and writer. This PR refactors the document source data reader to avoid a triple sequential read we had in the SBOM stream. I've dropped the original test as now we only need to check the correct assignment of the source data.

This PR also adds the computation of the stream hashes in sha1 (for spdx 2.x references) and sha512 .

We now also have a new `TrackSource` option in the unserializer to enable/disable the tracking of the source doc data. It is enabled by default.

### DIFF
--- a/pkg/reader/reader.go
+++ b/pkg/reader/reader.go
@@ -6,9 +6,11 @@ package reader
 //go:generate go run github.com/maxbrunsfeld/counterfeiter/v6 -generate
 
 import (
+	"crypto/sha1" //nolint:gosec // SHA1 is required in SPDX2
 	"crypto/sha256"
-	"encoding/hex"
+	"crypto/sha512"
 	"fmt"
+	"hash"
 	"io"
 	"os"
 	"sync"
@@ -128,6 +130,17 @@ func (r *Reader) ParseFileWithOptions(path string, o *Options) (*sbom.Document, 
 	return doc, nil
 }
 
+type byteCounter struct {
+	Count *int64
+}
+
+// Write implements the io.Writer interface, summing the bytes it gets.
+func (bc byteCounter) Write(p []byte) (int, error) {
+	n := len(p)
+	*bc.Count += int64(n)
+	return n, nil
+}
+
 // ParseStreamWithOptions returns a document from a ioreader, accept options for unserializer
 func (r *Reader) ParseStreamWithOptions(f io.ReadSeeker, o *Options) (*sbom.Document, error) {
 	if o == nil {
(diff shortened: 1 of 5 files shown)


puerco:
Wow this test really was a trip to debug. Added a note to the test file to warn other dwellers of all the fun you get when you mess with `init()`.

pkg/reader/reader.go
```diff
@@ -145,26 +160,60 @@ func (r *Reader) ParseStreamWithOptions(f io.ReadSeeker, o *Options) (*sbom.Docu
...
+	if o.UnserializeOptions.TrackSource {
+		doc.Metadata.SourceData = &sbom.SourceData{
+			Format: string(format),
+			Size:   *counter.Count,
+			Hashes: map[int32]string{},
+		}
+		for algo, hasher := range hashers {
+			doc.Metadata.SourceData.Hashes[int32(algo)] = fmt.Sprintf("%x", hasher.Sum(nil))
 		}
 	}
```
jhoward-lm, line 218: Is this capturing the actual hash of the source data, or always computing the hash of `nil`? Based on my comment above, if using `*bytes.Buffer`:

```suggestion
	if o.UnserializeOptions.TrackSource {
		doc.Metadata.SourceData = &sbom.SourceData{
			Format: string(format),
			Size:   buf.Cap(),
			Hashes: map[int32]string{},
		}

		data := buf.Bytes()

		for algo, hasher := range hashers {
			doc.Metadata.SourceData.Hashes[int32(algo)] = fmt.Sprintf("%x", hasher.Sum(data))
		}
	}
```

pkg/reader/reader.go
```diff
@@ -120,14 +124,25 @@ func (r *Reader) ParseFileWithOptions(path string, o *Options) (*sbom.Document,
...
+type byteCounter struct {
+	Count *int64
+}
+
+// Write implements the io.Writer interface, summing the bytes it gets.
+func (bc byteCounter) Write(p []byte) (int, error) {
+	n := len(p)
+	*bc.Count += int64(n)
+	return n, nil
+}
```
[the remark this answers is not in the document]
puerco: Ah so the problem with `bytes.Buffer` is that it's intended to be a buffer. You collect data in the buffer but you then are supposed to dump it elsewhere. This means that to use the buffer, we need to hold the hold sbom in memory and then count the length (or the allocated slice at least). The custom type just implemented io.Write and discarded data as it came in after counting it.

But we can try! I've pushed a new commit replacing the custom type with the buffer. We can revert it if we need to.

pkg/reader/reader.go
puerco: Ah no, we don't need to pass the data to the `Sum()` method because the hash.Hash interface implements io.Writer. the hashers are getting the SBOM through the `Write()` method as are the rest of the listeners.

puerco: (The test has an example SBOM and their hashes if you want to verify)

jhoward-lm: Oh ok, thanks for the info!

puerco:
Aaargh I'm such a dummy I didn't check the linter, it is fixed now. PTAL if you have a chance @jhoward-lm

jhoward-lm (approved):

PR #281, opened 2024-10-31
fix: encode source data hash as string

ashearin:
@jhoward-lm ran into issues using the new hash field in protobom/storage. This updates `setSourceData` to encode the hash as a string. Test values also updated.


jhoward-lm (approved):

cpanato (approved):
thanks

PR #306, opened 2025-01-23
fix: assign `PackageOriginator` under `Originators`

### LINKED ISSUE #251: Supplier overwrites Originator in SPDX output
Should https://github.com/protobom/protobom/blob/main/pkg/native/serializers/serializer_spdx23.go#L346-L353 really write the `p.PackageSupplier` or is the intention to assign to `p.PackageOriginator`?

Happy to send a PR to change this.

Vyom-Yadav:
Fixes #251


puerco (approved):
Awesome Vyom, something to note is that this change should have triggered a conformance error. But most likely our test SBOMs don't have the originator set.

I'm merging this as is, if you have an SBOM with the originator set we can add it to the tests.

Thanks!

PR #318, opened 2025-02-18
use go1.24 and update golangci-lint

cpanato:


puerco (approved):
Thank you, I like this approach more. We should probably codify it in the contribution guide at some point.

PR #324, opened 2025-02-26
[Fix]: replace invalid or empty serial id with newly generated valid one

### LINKED ISSUE #321: CycloneDX `serialNumber` Not Assigned UUID When Empty or Invalid
## Description

When converting an SPDX SBOM to CycloneDX using protobom@v0.5.1 (e.g., with sbom-convert), the `serialNumber` in the output isn’t always set properly. If the original SBOM doesn’t have a valid unique identifier or  empty, the converted CycloneDX file fails validation in Dependency-Track v4.12.5 because it expects a specific format (urn:uuid:...).

#### To Reproduce
- Convert an SPDX SBOM with no unique identifier to CycloneDX using sbom-convert.
- Upload the result to Dependency-Track.
- It fails with an error about serialNumber not matching the required format.

#### Expected Behavior
- The serialNumber should always be a valid unique identifier (like urn:uuid:...) in the CycloneDX output, even if the input SBOM doesn’t provide one.

viveksahu26:
closes #321 

This PR updates the following changes:
- It check whether SerialNumber is empty or invalid. In both cases, it adds a  newly generated UUID.

### DIFF
--- a/pkg/native/serializers/serializer_cdx.go
+++ b/pkg/native/serializers/serializer_cdx.go
@@ -5,10 +5,12 @@ import (
 	"errors"
 	"fmt"
 	"io"
+	"regexp"
 	"strconv"
 	"strings"
 
 	cdx "github.com/CycloneDX/cyclonedx-go"
+	"github.com/google/uuid"
 	cdxformats "github.com/protobom/protobom/pkg/formats/cyclonedx"
 	"github.com/protobom/protobom/pkg/native"
 	"github.com/protobom/protobom/pkg/sbom"
@@ -16,6 +18,9 @@ import (
 )
 
 var _ native.Serializer = &CDX{}
+
+// Precompiled regex for serialNumber validation
+var serialNumberPattern = regexp.MustCompile(`^urn:uuid:[0-9a-f]{8}-[0-9a-f]{4}-[0-9a-f]{4}-[0-9a-f]{4}-[0-9a-f]{12}$`)
 
 const (
 	stateKey state = "cyclonedx_serializer_state"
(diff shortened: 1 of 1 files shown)


pkg/native/serializers/serializer_cdx.go
```diff
@@ -149,6 +155,13 @@ func (s *CDX) Serialize(bom *sbom.Document, _ *native.SerializeOptions, _ interf
 	return doc, nil
 }
 
+// isValidCycloneDXSerialNumber validates serial id against regex pattern
+func isValidCycloneDXSerialNumber(serial string) bool {
+	pattern := `^urn:uuid:[0-9a-f]{8}-[0-9a-f]{4}-[0-9a-f]{4}-[0-9a-f]{4}-[0-9a-f]{12}$`
+	matched, _ := regexp.MatchString(pattern, serial)
```
puerco, line 161: Define the regexp outside of the function and make sure we compile it only once when needed as dointg it at every function call is expensive.

pkg/native/serializers/serializer_cdx.go
```diff
@@ -45,6 +47,10 @@ func (s *CDX) Serialize(bom *sbom.Document, _ *native.SerializeOptions, _ interf
 
 	doc := cdx.NewBOM()
 	doc.SerialNumber = bom.Metadata.Id
+	if doc.SerialNumber == "" || !isValidCycloneDXSerialNumber(doc.SerialNumber) {
+		doc.SerialNumber = "urn:uuid:" + uuid.New().String()
```
puerco, line 51: Lets use here a deterministic algorithm to generate the new serial number. When the protobom document has an ID defined, use an uuid sha1 to generate the uuid, something like:

```golang
uuid.NewSHA1(uuid.MustParse(`5dbcd03c-dd56-4fff-97af-77f89c66eeba`), []byte(protobom id))
```

puerco (commented):
Thanks for fixing this @viveksahu26  ! A few initial comments :point_down:

pkg/native/serializers/serializer_cdx.go
puerco, line 51: Since this is essentially altering the protobom data, we have an option to fail the conversion when the protobom document does not have an Id set. 

See how the SPDX serializer implements it here and make sure we honor it here too:
https://github.com/protobom/protobom/blob/5612e368b95d757401f322ed4939b3957d8d40a0/pkg/native/serializers/serializer_spdx23.go#L88-L95

pkg/native/serializers/serializer_cdx.go
```diff
@@ -17,22 +19,27 @@ import (
...
 	}
 )
 
 func NewCDX(version, encoding string) *CDX {
 	return &CDX{
-		version:  version,
-		encoding: encoding,
+		version:              version,
+		encoding:             encoding,
+		GenerateSerialNumber: false,
```
puerco, line 42: Mmh no, sorry... Now that I look at it, the cdx options are not initialized in the serializer. Sorry for the wrong comment in the previous review. 

Let me open a PR to set them up so that you can use them here.

puerco: OK, here: https://github.com/protobom/protobom/pull/327

Feel free to cherry-pick that commit if it helps unblocking you (and I'll close 327).

puerco: Awesome, we merged #327. Just rebase and use the new option to control the serial number generation.

viveksahu26: done !!

puerco (approved):
Awesome, thanks @viveksahu26 !

PR #325, opened 2025-02-26
set version to 1 from 0

### LINKED ISSUE #322: CycloneDX version Set to 0 Instead of 1 After Conversion
## Description

When converting an SBOM from SPDX to CycloneDX using protobom@v0.5.1 (e.g., with `sbom-convert`), the version field in the output CycloneDX file is set to `0` if no version is specified in the input. Dependency-Track v4.12.5 rejects this because it requires the version to be at least `1`.

#### To Reproduce
- Convert an SPDX SBOM with no version info to CycloneDX using sbom-convert.
- Upload the result to Dependency-Track.
- It fails with an error saying the version must be at least 1.

#### Expected behavior
- The version in the CycloneDX output should default to 1.

viveksahu26:
closes #322 

This PR adds the following changes:
- It updated the version value from 0 to 1


pkg/native/serializers/serializer_cdx.go
```diff
@@ -47,7 +47,9 @@ func (s *CDX) Serialize(bom *sbom.Document, _ *native.SerializeOptions, _ interf
 	doc.SerialNumber = bom.Metadata.Id
 	ver, err := strconv.Atoi(bom.Metadata.Version)
 	// TODO(deprecation): If version does not parse to int, there's data loss here.
-	if err == nil {
+	if err != nil || ver < 1 {
+		doc.Version = 1 // Default to 1
+	} else {
```
puerco, line 52: Since the CDX serializer is the only user of the version field, change the init of the version number from 0 to 1 here:

https://github.com/protobom/protobom/blob/5612e368b95d757401f322ed4939b3957d8d40a0/pkg/sbom/document.go#L26

If we get a zero here it's fine to bump it to 1

puerco (commented):
Thanks for the fix, just an additional detail to change.

(Note that some of these changes may require you to regenerate the conformance data)

puerco (approved):
Thanks @viveksahu26 I pushed a commit to your branch with the updated conformance data and the tests are now passing.  

Thank you! :rocket:

PR #328, opened 2025-02-27
skip files section with noassert license

viveksahu26:
closes https://github.com/protobom/protobom/issues/323

This PR updates the following changes:
- This change updates the unserialization process—when reading SPDX files, it filters out "NOASSERTION" entries from licenseInfoInFiles and licenseConcluded, like it does for SPDX packages, ensuring the output CycloneDX SBOM only carries valid licenses.


puerco (approved):
LGTM, it was again failing because the changes make the conformance data to fail. I pushed a commit regenerating the representation of the curl SBOM that included some `NOASSERTION` entries in the files. Wehenever that happens and it is expected (as in this case) just run `make conformance` and check-in the results

We should probably add a test to ensure we are filtering all the `NOASSERTIONS` correctly, but let's get this in and we can do it in a follow up.

Thanks @viveksahu26 !

viveksahu26:
Thank  you soo much for handling conformance data failure.

PR #332, opened 2025-03-04
Fix 1-char bug flattening CDX data.

puerco:
This PR fixes an embarrassing 1-character bug that was causing the CDX boms to be flattened when read. 

The fix is on one commit, the other is just the regenerated conformance data.

Signed-off-by: Adolfo Garcia Veytia (puerco) <puerco@carabiner.dev>


cpanato (approved):

cpanato:
Thanks

PR #339, opened 2025-04-07
Update CDX (un)serializer to handle timestamp.

### LINKED ISSUE #335: Timestamp for CDX unseralizer
It looks to me like we are not getting the timestamp from CDX:

https://github.com/protobom/protobom/blob/8c5ee7c0f5d2f448115c3243e971815c9ffb917f/pkg/native/unserializers/unserializer_cdx.go#L48

This could be set to bom.Metadata.Timestamp
https://github.com/CycloneDX/cyclonedx-go/blob/v0.9.2/cyclonedx.go#L962

I guess the string needs to be decoded and converted to a timestamppb.

Am I missing something?

jeffmendoza:
Fixes #335


puerco (approved):
Thanks Jeff!!!

PR #346, opened 2025-07-10
create id for root node if that is empty

cpanato:
Fixes: https://github.com/protobom/protobom/issues/329

or try to

not sure if is that what you want/need @puerco

### DIFF
--- a/pkg/sbom/nodelist.go
+++ b/pkg/sbom/nodelist.go
@@ -3,11 +3,11 @@ package sbom
 import (
 	"fmt"
 	"reflect"
+	"slices"
 	"sort"
 	"strings"
 
 	"github.com/google/go-cmp/cmp"
-	"github.com/sirupsen/logrus"
 )
 
 // This file adds a few methods to the NodeList type which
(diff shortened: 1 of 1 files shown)


pkg/sbom/nodelist.go
```diff
@@ -193,13 +194,13 @@ func (nl *NodeList) MergeEdges(es []*Edge) {
 // More than one root element can be added to the NodeList.
 func (nl *NodeList) AddRootNode(n *Node) {
 	if n.Id == "" {
-		// TODO warn here
-		return
+		logrus.Warnf("Node with empty ID, creating a new Id: %s@%s\n", n.Name, n.Version)
+		n.Id = NewNodeIdentifier(fmt.Sprintf("%s@%s", n.Name, n.Version))
```
puerco, line 198: Cool, yes autogenerate the identifier, but keep the  `auto` prefix so that the library knows it was added internally. Also drop the logrus calls, we have been trying to remove them:

```suggestion
		n.Id = NewNodeIdentifier("auto", fmt.Sprintf("%s@%s", n.Name, n.Version))
```

pkg/sbom/nodelist.go
```diff
@@ -193,13 +194,13 @@ func (nl *NodeList) MergeEdges(es []*Edge) {
...
 	}
 
 	for _, id := range nl.RootElements {
 		if id == n.Id {
-			// TODO warn here
+			logrus.Warnf("Node with ID %s already exists in the NodeList, skipping: %s\n", n.Id, n.Name)
```
puerco, line 203: Just return here, i think it's fine if we keep this as a noop. IMHO adding a node as root node should be idempotent.
```suggestion
```

pkg/sbom/nodelist.go
```diff
@@ -193,13 +194,13 @@ func (nl *NodeList) MergeEdges(es []*Edge) {
 // More than one root element can be added to the NodeList.
 func (nl *NodeList) AddRootNode(n *Node) {
 	if n.Id == "" {
-		// TODO warn here
-		return
+		logrus.Warnf("Node with empty ID, creating a new Id: %s@%s\n", n.Name, n.Version)
+		n.Id = NewNodeIdentifier(fmt.Sprintf("%s@%s", n.Name, n.Version))
 	}
 
 	for _, id := range nl.RootElements {
```
puerco, line 201: We can now change this loop to use the slices package
```suggestion
	if slices.Contains(nl.RootElements, id) {
	   return
	}
```

puerco (commented):
Thanks! just a nit on the fix it, and let's just drop the logrus calls I think we don't need them as it can "just work":

pkg/sbom/nodelist.go
cpanato: got it

Issue #353, opened 2025-08-12
Example3 in README.md doesn't work


KAWAHARA-souta:
I've got following error when execute Example3 in README.md

```
$ go run README_example3.go 
# command-line-arguments
./README_example3.go:77:11: document.RelateNodeAtID undefined (type *sbom.Document has no field or method RelateNodeAtID)
./README_example3.go:88:11: document.RelateNodeAtID undefined (type *sbom.Document has no field or method RelateNodeAtID)
```

We should fix like below:
```
-       document.RelateNodeAtID(node1, appNode.Id, sbom.Edge_contains)
-       document.RelateNodeAtID(node2, appNode.Id, sbom.Edge_contains)
+       document.NodeList.RelateNodeAtID(node1, appNode.Id, sbom.Edge_contains)
+       document.NodeList.RelateNodeAtID(node2, appNode.Id, sbom.Edge_contains)
```

PR #354, opened 2025-08-12
Fix example3 in README.md

KAWAHARA-souta:
Resolves: #353


puerco (approved):
Thanks @KAWAHARA-souta !

PR #356, opened 2025-08-13
Fix release process

paulinayanez:
We attempt to fix the release process by flipping the checkout and go setup order in release.


puerco (approved):
Thanks!

PR #370, opened 2025-10-10
feat: Implement dup & self destination filter in RelateNodeAtID

KAWAHARA-souta:
RelateNodeAtID() was adding relationships without filtering for duplicates, so I implemented filter.
Additionally, I implemented filtering for self-relationships since I deemed them unnecessary.

I also added and updated unit tests for those.

### DIFF
--- a/pkg/sbom/nodelist.go
+++ b/pkg/sbom/nodelist.go
@@ -626,6 +626,11 @@ func (nl *NodeList) RelateNodeAtID(n *Node, nodeID string, edgeType Edge_Type) e
 		}
 	}
 
+	// Check if self destination, then early return
+	if nodeID == n.Id {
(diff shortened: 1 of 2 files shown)


pkg/sbom/edge.go
```diff
@@ -62,3 +58,12 @@ func (e *Edge) AddDestinationById(ids ...string) {
...
 }
+
+// indexDestinations returns a set with the IDs of the destinations
+func (e *Edge) indexDestinations() map[string]struct{} {
+	dests := map[string]struct{}{}
+	for _, id := range e.To {
+		dests[id] = struct{}{}
+	}
+	return dests
+}
```
puerco: Love it, thanks!

PR #387, opened 2026-01-12
Actually fix cdx regex init data race

wlynch:
This is a follow up to https://github.com/protobom/protobom/pull/386, which didn't actually fix the problem w.r.t. the race detector because of the unsynchronized read.

This removes the read, since it doesn't really make a difference for the init once behavior.


cpanato (approved):

cpanato:
thanks!

PR #407, opened 2026-04-20
Add support for filter out dep types

puerco:
This adds a new function to the nodelist: `RemoveNodesByEdgeType()`

This new function allow to remove entire classes of nodes depending on how they are connected to the graph. The utility here is that we can now remove all the test dependencies or build dependencies from an SBOM.

RemoveNodesByEdgeType removes all nodes connected by the specified edge type. When a node has more than one path, it remains but it loses the targeted edge.


puerco:
I've pushed four more commits fixing a bunch of bugs I just found.

cpanato (approved):
Thanks

PR #410, opened 2026-04-22
Update SPDX media type strings

### LINKED ISSUE #409: Update IANA media type strings of SPDX formats
The media type strings of SPDX formats here:

https://github.com/protobom/protobom/blob/fd106f299866343530192a87129cc8fe19de89f1/pkg/formats/formats.go#L16-L19

are slightly different from information on IANA where it uses [`application/spdx+json`](https://www.iana.org/assignments/media-types/application/spdx+json) for SPDX JSON format and [`text/spdx`](https://www.iana.org/assignments/media-types/text/spdx) for SPDX Tag:value format.

Proposes change:

```go
	SPDX23TV   = Format("text/spdx;version=2.3")
	SPDX23JSON = Format("application/spdx+json;version=2.3")
	SPDX22TV   = Format("text/spdx;version=2.2")
	SPDX22JSON = Format("application/spdx+json;version=2.2")
```

Note that since the media type strings for `SPDX*TV` still contatin "text" (as a prefix), this function should still work:

https://github.com/protobom/protobom/blob/fd106f299866343530192a87129cc8fe19de89f1/pkg/formats/formats.go#L76-L87

bact:
Use media type strings as registered with IANA (`application/spdx+json` and `text/spdx`):

- https://www.iana.org/assignments/media-types/application/spdx+json
- https://www.iana.org/assignments/media-types/text/spdx

`Encoding()` and `Type()` should still working as before, since they are rely on strings containing "text", "json", and "spdx", which the new strings are still contain.

To fix #409

### DIFF
--- a/pkg/formats/formats.go
+++ b/pkg/formats/formats.go
@@ -13,10 +13,10 @@ const (
 	JSON       = "json"
 	XML        = "xml"
 	TEXT       = "text"
-	SPDX23TV   = Format("text/spdx+text;version=2.3")
-	SPDX23JSON = Format("text/spdx+json;version=2.3")
-	SPDX22TV   = Format("text/spdx+text;version=2.2")
-	SPDX22JSON = Format("text/spdx+json;version=2.2")
+	SPDX23TV   = Format("text/spdx;version=2.3")
+	SPDX23JSON = Format("application/spdx+json;version=2.3")
+	SPDX22TV   = Format("text/spdx;version=2.2")
+	SPDX22JSON = Format("application/spdx+json;version=2.2")
 	CDX10JSON  = Format("application/vnd.cyclonedx+json;version=1.0")
 	CDX11JSON  = Format("application/vnd.cyclonedx+json;version=1.1")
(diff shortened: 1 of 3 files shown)


puerco:
Thanks for looking into this @bact . These flags are not intended to be compatible with anything, they are protobom-only constants we use to specify the formats. While it would have been useful initially to define them according to a more canonical source, they are not intended to be compatible with other systems that use the standardized strings.

Unfortunately, we cannot change them now without breaking users because anything not using the exposed constants will fail, especially any CLI tools that pass the format string will break.

I will close this now but happy to keep chatting about it.

bact:
Thank you for the explanation. As internal constants, it is understandable.

(I think it is not a breaking change, based on current string-matching way that `Encoding()` and `Type()` do. But of course, not touching them at all is safer.)

Btw, SPDX 3 JSON now has an official media type: `application/spdx3+json`.
https://www.iana.org/assignments/media-types/application/spdx3+json 

As pkg/native/serializers/beta/serializer_spdx3.go is still in beta, may be we can change only the SPDX 3 string :)

PR #419, opened 2026-06-04
Drop 1.(6>7) serializiation hack

puerco:
This PR removes the 1.7 hack we added in #406 to support early cyclonedx 1.7 and we now support 1.7 with the native library.

Supersedes #414 

/cc @cpanato


puerco:
@cpanato  Sorry for the large diff, I bumped the linter and fixed all the new nits but the real changes are in 9d328f8

cpanato (approved):

PR #420, opened 2026-06-04
Update CI

puerco:
This PR updates a couple of things in the project's CI:

- We drop the stale issue workflow as it has caused more trouble than help
- SLSA Source workflows are now pinned
- Tests are now run in matrix on both supported go versions and on all supported platforms
- Finally the SBOM generation has now been updated to use unpack fore a more detailed SBOM.

Signed-off-by: Adolfo García Veytia (Puerco) <puerco@carabiner.dev>


puerco:
Ah it needs #419 and a rebase :/

cpanato (changes requested):
you removed the .go-version, need to fix the golangci-lint job

puerco:
I've rebased and pushed it and it's now working. The new matrix testing also uncovered a bug on windows, fixed it too.

cpanato (approved):

PR #421, opened 2026-06-05
fix(cdx): surface metadata.tools in the CycloneDX unserializer

### LINKED ISSUE #417: CycloneDX unserializer drops `metadata.tools` — `Document.Metadata.Tools` always empty
Disclosing up-front that I used Claude Code to generate this issue...

### Summary
When parsing a CycloneDX SBOM, `d.GetMetadata().GetTools()` always returns an empty slice, even when the source document has a populated `metadata.tools`. The SPDX unserializer surfaces creators correctly, so there's a format asymmetry: consumers who want to know which tool generated an SBOM have to fall back to parsing the raw blob themselves for CDX.

### Version
protobom v0.5.5 (also present in v0.5.4; based on a quick read of `main`, still present there).

### Where
`pkg/native/unserializers/unserializer_cdx.go:50` initializes the metadata with `Tools: []*sbom.Tool{}` and nothing downstream ever populates it from the CycloneDX document's `metadata.tools`.

### Reproduction
Parse any CycloneDX SBOM that has `metadata.tools` populated — for example one produced by `cyclonedx-gomod`, `syft`, `trivy`, or `mikebom`:

```go
r := reader.New()
d, _ := r.ParseStream(f) // CycloneDX JSON
fmt.Println(len(d.GetMetadata().GetTools())) // 0, regardless of input
```

This happens for both shapes the spec allows:
- **CDX 1.4** legacy array: `metadata.tools: [ { vendor, name, version }, ... ]`
- **CDX 1.5+** object form: `metadata.tools: { components: [...], services: [...] }`

cyclonedx-gomod still emits the legacy array form under spec 1.6, so both shapes are live in the wild and both need to be handled on read.

### Expected
`d.GetMetadata().GetTools()` returns one `*sbom.Tool` per entry, with `Name`, `Version`, and `Vendor` populated.

### Notes on the 1.5+ object form
There's no single `vendor` field on a tool component. Real generators populate different fields:
- mikebom → `publisher`
- trivy → `manufacturer.name`
- syft → `author`
- aquasecurity (group on trivy components) → `group`

A reasonable fallback order is publisher → manufacturer.name → author → group. Happy to send a PR if the approach sounds right.

### Why it matters
Downstream consumers (we're using protobom in a supply-chain platform) need the tool name/version to drive ingestion behavior, attribution, and per-generator quirks. Today this requires re-parsing the raw blob alongside protobom's parse, which defeats the purpose of using a normalized SBOM library.

### Related
- #68 (open) proposes deprecating `Tool` for `Node`. Whether or not that lands, the CDX unserializer should populate *something* — today it silently drops the data.
- #305 (closed) catalogs data-loss points but doesn't list this one.

arpitjain099:
## What this fixes

Fixes #417.

The CycloneDX unserializer was dropping `metadata.tools`. After parsing any CDX SBOM, `Document.Metadata.Tools` was always empty, even when the source document had a populated tool list. The metadata struct is initialized with `Tools: []*sbom.Tool{}` and nothing downstream ever read `bom.Metadata.Tools`, so the data was silently lost on read.

This also closes a format asymmetry. The SPDX unserializer already surfaces its `Tool` creators from `CreationInfo`, so a consumer that wants to know which tool generated an SBOM gets it for SPDX but not for CycloneDX. With this change CDX behaves the same way.

## Root cause

In `pkg/native/unserializers/unserializer_cdx.go`, the `if bom.Metadata != nil` block handled lifecycles, the root component, and the timestamp, but never touched `bom.Metadata.Tools`. The field was initialized empty and left empty.

## Reproduction (before the fix)

```go
r := reader.New()
d, _ := r.ParseStream(f) // a CycloneDX JSON with metadata.tools populated
fmt.Println(len(d.GetMetadata().GetTools())) // prints 0 regardless of input
```

Confirmed against both shapes the spec allows, using a minimal CDX document for each:

- CDX 1.4 legacy array: `metadata.tools: [ { vendor, name, version }, ... ]` parsed to 0 tools.
- CDX 1.5+ object form: `metadata.tools: { components: [...] }` parsed to 0 tools.

## The fix

A new `metadataToolsToTools` helper maps `bom.Metadata.Tools` (a `cdx.ToolsChoice`) into protobom `Tool` entries, and it is now called in the metadata block. It handles every shape `ToolsChoice` can carry:

- The legacy `Tools` array (`[]cdx.Tool`), which is still emitted by generators like `cyclonedx-gomod` even under spec 1.6, so it stays relevant on read.
- The 1.5+ `Components` form. A tool component has no single vendor field, so the vendor is resolved through a fallback chain that matches what real generators populate: `publisher`, then `manufacturer.name`, then the deprecated `author` string, then `group`, then `supplier.name`.
- The 1.5+ `Services` form, using `provider.name` (or `group`) as the vendor analog.

Entries without a usable name are skipped. Nothing on the serialization path or any other field is touched.

## After the fix

Both shapes now surface `Name`, `Version`, and `Vendor`:

- Legacy array `{ "vendor": "anchore", "name": "syft", "version": "0.100.0" }` -> one `Tool{Name: "syft", Version: "0.100.0", Vendor: "anchore"}`.
- Components form `{ "publisher": "anchore", "name": "syft", "version": "1.0.0" }` -> one `Tool{Name: "syft", Version: "1.0.0", Vendor: "anchore"}`.

## Tests

- `TestMetadataToolsToTools` is a table test over the helper: nil choice, legacy array (including an all-empty entry that is skipped), the components form with `publisher`, the components vendor fallbacks (`manufacturer.name`, `author`, `group`), a nameless component that is skipped, and the services form.
- `TestUnserializeMetadataTools` is an end-to-end regression guard that runs the full `Unserialize` path on a real CDX JSON for both the legacy array and the components object form and asserts the tools are surfaced.

I verified the end-to-end test fails (0 tools, the exact bug symptom) when the new call is removed and passes with it in place. `go build ./...`, `go test ./pkg/native/... ./test/...` (including the conformance suite), `go vet`, and `gofmt`/`gofumpt` are all clean.

## Notes

- No proto change is needed; `sbom.Tool` already has `Name`, `Version`, and `Vendor`.
- This is read-only behavior for the unserializer. The existing serializer write path (`ToolsChoice.Tools`) is unchanged.

### DIFF
--- a/pkg/native/unserializers/unserializer_cdx.go
+++ b/pkg/native/unserializers/unserializer_cdx.go
@@ -93,6 +93,7 @@ func (u *CDX) Unserialize(r io.Reader, _ *native.UnserializeOptions, _ interface
 				md.Date = timestamppb.New(t)
 			}
 		}
+		md.Tools = append(md.Tools, u.metadataToolsToTools(bom.Metadata.Tools)...)
 	}
 
 	// Cycle all components and get their graph fragments. A CycloneDX BOM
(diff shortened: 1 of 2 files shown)


puerco (changes requested):
Mmh thanks for looking into this @arpitjain099 but I'm afraid this is not correct. Beyond the fact that it is truncating the tools tree, the problem is a little more complex than this:

The issue with the CDX tool section is that it introduces a full separate component tree. This is a full featured tree equivalent to the main graph: Components can have descendants and can reference elements in the main tree. This PR is essentialy cropping the tools tree at the first level and destroying the component data.

Before solving this problem please draft a technical proposal on the issue so that we can discuss it. Any PR solving it will require both the serializer and unserializer implementations as well as round tripping tests.

arpitjain099:
> Mmh thanks for looking into this @arpitjain099 but I'm afraid this is not correct. Beyond the fact that it is truncating the tools tree, the problem is a little more complex than this:
> 
> The issue with the CDX tool section is that it introduces a full separate component tree. This is a full featured tree equivalent to the main graph: Components can have descendants and can reference elements in the main tree. This PR is essentialy cropping the tools tree at the first level and destroying the component data.
> 
> Before solving this problem please draft a technical proposal on the issue so that we can discuss it. Any PR solving it will require both the serializer and unserializer implementations as well as round tripping tests.

Hello @puerco I see. Let me check the PR later tonight.

arpitjain099:
You're right, thanks for catching that. The flat `Tool{Name, Version, Vendor}` mapping is lossy for the CDX 1.5+ `metadata.tools.components` form, where a tool component can carry hashes, external references, properties, and even nested components, none of which survive that shape. I'd rather get the model right than land a partial conversion, so I'll close this and move the discussion to #417 with a proposal. The current branch stays on my fork for reference once we agree on the approach.

PR #426, opened 2026-07-01
perf(sbom): make NodeList graph traversal O(nodes+edges)

### LINKED ISSUE #425: `NodeList` graph lookups are O(nodes) linear scans, making traversal/consumer code O(nodes²)
**Note:** This was all generated by Claude Code for me based on an investigation I have it perform of some slow SBOM parsing instances. I don't really understand the internals of protobom very well at all myself.

## Summary

`NodeList.GetNodeByID` resolves a node by ID with a **linear scan** of the entire
`Nodes` slice:

```go
// pkg/sbom/nodelist.go (as of v0.5.6)
func (nl *NodeList) GetNodeByID(id string) *Node {
	for i := range nl.Nodes {
		if nl.Nodes[i].Id == id {
			return nl.Nodes[i]
		}
	}
	return nil
}
```

That is fine for a single lookup, but several protobom methods call it (directly or
transitively) **once per graph edge / per visited node**, which turns whole-graph
operations into **O(nodes²)** (or O(nodes·edges)). On large SBOMs — e.g. container-image
SBOMs with 100k–200k nodes — this is minutes of CPU.

## Affected methods (all in `pkg/sbom/nodelist.go`)

There are **two** independent quadratic sources:

**A. `GetNodeByID` linear scans in loops**

- **`NodeSiblings(id)`** — scans **all** edges (`for _, r := range nl.Edges`) and calls
  `nl.GetNodeByID(to)` for each edge target. O(edges) per call on its own.
- **`indexConnectedNodes` / `connectedIndexRecursion`** — calls `NodeSiblings(s.Id)` for
  every node reached in the recursion, so it compounds to O(nodes·edges).
- **`NodeDescendants(id, maxDepth)`** — calls `nl.GetNodeByID(siblingID)` inside the
  nested edge-traversal loops.

**B. `RelateNodeAtID` rebuilds all indexes on every call**

`RelateNodeAtID` calls both `nl.indexNodes()` and `nl.indexEdges()` (each O(nodes+edges))
at the top of every invocation. `NodeDescendants` calls it once per descendant in its
final loop, making that loop **O(descendants·(nodes+edges))**. In practice this is the
*dominant* cost for `NodeDescendants` — fixing only the `GetNodeByID` call above barely
moved its runtime; the per-call index rebuilds did. Any consumer that calls
`RelateNodeAtID` in a loop hits the same wall.

The single-shot linear scanners `GetNodeByID`, `GetNodesByName`, and
`GetNodesByIdentifier` are individually fine, but they are easy to misuse inside a loop —
a downstream consumer that calls `GetNodeByID` per edge (a very natural thing to do when
walking the dependency graph) inherits the same O(nodes²) blow-up without any obvious
signal that it's happening.

## Reproduction / measurement

Build a synthetic flat SBOM with N package nodes and ~N dependency edges (root depends on
all N, plus a chain), then call `NodeDescendants(root, N+1)`:

| Nodes | Before   | After  |
|------:|---------:|-------:|
| 20k   | ~19.7 s  | ~16 ms |
| 40k   | ~1m 21s  | ~23 ms |

Building the id→node map once and batching the ancestor-edge construction (see the
companion PR) drops `NodeDescendants` from ~20 s to ~16 ms at 20k nodes, and makes the
curve linear instead of quadratic.

## Impact

Any protobom consumer that walks the dependency graph and resolves nodes by ID (a common
pattern) sees multi-minute processing on large SBOMs. In our case this exceeded a
15-minute processing deadline and caused ingestion failures; we worked around it in our
own code by building an ID→index map once, but the same fix applied upstream would help
`NodeDescendants`/`NodeSiblings`/`indexConnectedNodes` and every downstream user.

## Proposed fix

Two independent, low-risk changes:

1. **Fix protobom's own quadratic callers.** In `NodeDescendants`,
   `NodeSiblings`, and the `connectedIndexRecursion` path, build the node index once (via
   the existing `indexNodes()` helper, and the existing `indexEdges()` for the all-edges
   scan) and look up in O(1) instead of calling `GetNodeByID` per node/edge. This mirrors
   the pattern already used elsewhere in the file (e.g. `cleanEdges`, the top of
   `NodeDescendants` already builds `indexEdges()`/`indexRootElements()`).

2. **Document `GetNodeByID` as O(n).** Add a doc-comment noting it's a linear scan and
   that callers resolving many IDs should build an index once (e.g. `indexNodes()` if it
   were exported, or their own map). Optionally, expose a batch helper such as
   `GetNodesByID(ids []string) map[string]*Node` so consumers have an O(nodes+lookups)
   path without reaching for internals.

A cached lazily-built index on `NodeList` itself was considered, but `NodeList` is a
generated protobuf message and callers mutate `Nodes`/`Edges` directly, so a cache would
risk going stale; building the index at the start of each traversal is safe and matches
existing code.

See the companion PR write-up for concrete before/after diffs.

nchelluri:
Several NodeList graph operations were quadratic on large SBOMs, from two independent causes:

1. GetNodeByID is a linear scan, and NodeSiblings / indexConnectedNodes / NodeDescendants called it once per edge or per visited node — O(nodes^2).

2. RelateNodeAtID rebuilds the full node and edge indexes (indexNodes + indexEdges) on every call, and NodeDescendants called it once per descendant — O(descendants * (nodes+edges)). This was the dominant cost for NodeDescendants; fixing only the GetNodeByID call barely moved it.

Build the needed index(es) once and reuse them:

- NodeDescendants: build the node index once; replace the per-descendant RelateNodeAtID loop with a single Edge_ancestor edge built in one pass (same result the loop produced).
- indexConnectedNodes/connectedIndexRecursion: build node+edge indexes once and walk edgeIdx directly instead of calling NodeSiblings per node.
- NodeSiblings: build the node index once instead of GetNodeByID per target.
- Document GetNodeByID as an O(n) linear scan.

Behavior is unchanged; the full test suite (pkg/sbom, test/conformance, readers, serializers, storage, writer) passes. On a synthetic flat graph, NodeDescendants drops from ~1m21s to ~23ms at 40k nodes (~3500x) and the curve becomes linear.

Fixes #425

### DIFF
--- a/pkg/sbom/nodelist.go
+++ b/pkg/sbom/nodelist.go
@@ -488,7 +488,11 @@ func (nl *NodeList) GetNodesByName(name string) []*Node {
 	return ret
 }
 
-// GetNodeByID returns a node with the specified ID
+// GetNodeByID returns a node with the specified ID.
+//
+// This is a linear scan of the node list — O(number of nodes). Callers that
+// need to resolve many IDs (e.g. while walking the dependency graph) should
+// avoid calling this in a loop and instead build an id->node map once.
 func (nl *NodeList) GetNodeByID(id string) *Node {
 	for i := range nl.Nodes {
 		if nl.Nodes[i].Id == id {
@@ -873,7 +877,8 @@ func (nl *NodeList) NodeGraph(id string) *NodeList {
 // stop when reaching them.
 func (nl *NodeList) indexConnectedNodes(id string) nodeIndex {
 	index := nodeIndex{}
-	node := nl.GetNodeByID(id)
+	nodeIdx := nl.indexNodes()
+	node := nodeIdx[id]
 	if node == nil {
 		return index
 	}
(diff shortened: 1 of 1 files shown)


puerco (approved):
This is awesome, thanks @nchelluri ! I'll cut a release with the patch asap.

PR #428, opened 2026-07-01
perf(cdx): batch component merge so ParseStream is O(nodes+edges)

### LINKED ISSUE #427: CycloneDX `ParseStream` is O(N²) in the number of components
## Summary

Parsing a CycloneDX BOM is quadratic in the number of components. The unserializer relates
each component to the document **one at a time**, and the merge primitives it uses rebuild
a full index of the *growing* document on every call. A BOM with N components therefore
re-indexes the whole graph N times, so `reader.ParseStream` (via the CDX unserializer) is
**O(N²)**.

On large container-image SBOMs (~100k nodes) this is ~8 minutes of parse time. As of
v0.5.7 the code is unchanged, so this still reproduces on the latest release.

## Where

`pkg/native/unserializers/unserializer_cdx.go`, in `Unserialize`:

```go
if bom.Components != nil {
	for i := range *bom.Components {
		nl, err := u.componentToNodeList(&(*bom.Components)[i], &cc)
		if err != nil {
			return nil, fmt.Errorf("converting component to node: %w", err)
		}

		if !hasRootComponent {
			doc.NodeList.Add(nl)                                           // per component
		} else {
			if err := doc.NodeList.RelateNodeListAtID(
				nl, doc.NodeList.RootElements[0], sbom.Edge_contains); err != nil { // per component
				return nil, fmt.Errorf("relating components to root node: %w", err)
			}
		}
	}
}
```

Both merge primitives (in `pkg/sbom/nodelist.go`) rebuild indexes of the destination
NodeList on **every** call:

- `NodeList.Add` → `nl.indexNodes()`, `nl.MergeEdges(...)`, `nl.indexRootElements()`, and
  `nl.cleanEdges()` (which itself calls `indexNodes()` and rebuilds the edge slice).
- `NodeList.RelateNodeListAtID` → `nl.indexNodes()` + `nl.indexEdges()`.

The destination is `doc.NodeList`, which grows with each merged component. So call K
indexes ~K components' worth of nodes/edges, and summing over N components gives O(N²).

## Why CycloneDX and not SPDX

CycloneDX components are **nested** (a component may contain sub-components), so the
unserializer builds a NodeList *fragment* (subtree) per top-level component and merges each
fragment into the document — hitting the per-call reindex.

SPDX is **flat** (packages / files / relationships are top-level lists), so its unserializer
appends nodes and edges directly with `AddNode` / `AddEdge` (O(1) each). SPDX parsing is
already O(N) and does not have this problem.

## Reproduction / measurement

Build a synthetic flat CycloneDX 1.5 BOM with N components and a `metadata.component` root,
then parse it with the CDX unserializer (`Unserialize`). Measuring only parse, on current
`main` (v0.5.7):

| Components | `ParseStream` |
|----------:|--------------:|
| 20k | ~16.2 s |
| 40k | ~1m 04s |

Doubling the component count roughly quadruples the time — the quadratic signature.

## Impact

Any consumer that parses moderately large CycloneDX SBOMs (container images, monorepo
dependency graphs) pays minutes of CPU. In our case the parse exceeded a 15-minute
processing deadline and caused ingestion failures; we shipped the fix below as a vendored
fork in the meantime.

## Proposed fix

Accumulate every component's graph fragment into a single `NodeList` and merge it into the
document **once**, so the graph is indexed a single time → **O(nodes+edges)**. Node de-dup
by id is preserved (a seen-set replaces the de-dup that `Add` did internally). Parse output
is unchanged; the `pkg/native/unserializers` and `test/conformance` round-trip suites pass.

Measured after the fix: 20k → ~42 ms, 40k → ~56 ms (vs ~16.2 s / ~1m 04s before) — the
curve becomes linear.

A more general alternative would be to make `Add` / `RelateNodeListAtID` cheap to call
repeatedly (accept a prebuilt index, or drop the per-call `cleanEdges`); batching at the
call site is the smaller, lower-risk change and fixes the hot path.

See the companion PR: `perf(cdx): batch component merge so ParseStream is O(nodes+edges)`.

## Related

Separate from the `NodeList` graph-*traversal* O(N²) fixed in #426
(`NodeDescendants` / `NodeSiblings` / `indexConnectedNodes`) — that is a different code path
(graph queries, not parsing). This issue is specifically the CycloneDX unserialize/parse
path.

nchelluri:
## What

Make CycloneDX `ParseStream` linear in the size of the BOM.

The unserializer related each component to the document **one at a time**: for every
component it calls `NodeList.Add` (headless BOMs) or `NodeList.RelateNodeListAtID` (BOMs
with a `metadata.component` root). Both methods rebuild a full node+edge index of the
*growing* document on every call (`indexNodes` + `indexEdges`, and `Add` also runs
`cleanEdges`). So a BOM with N components re-indexes the whole graph N times — parsing is
**O(N²)**.

On large container-image SBOMs (~100k nodes) this is ~8 minutes of parse time.

## Fix

Accumulate every component's graph fragment into a single `NodeList` and merge it into the
document **once**, so the graph is indexed once → **O(nodes+edges)**. Node de-duplication
by id is preserved (`Add` already de-duped; the batched path keeps that with a seen-set).
Parse output is unchanged.

## Measured

Synthetic flat CycloneDX 1.5 BOM with a `metadata.component` root (the
`RelateNodeListAtID`-per-component path), timing the CDX unserializer's `Unserialize`.
Directly comparing `main` against this branch:

| Components | `main` (before) | this branch (after) | Speedup |
|----------:|----------------:|--------------------:|--------:|
| 20k | 16.2 s | 42 ms | ~385× |
| 40k | 1m 04s | 56 ms | ~1,150× |

On `main`, doubling the component count (20k→40k) roughly quadruples parse time
(quadratic); on this branch it stays flat (linear).

## Test plan
- [x] `go build ./...`, `go vet ./pkg/native/unserializers/`
- [x] `go test ./...` — full suite passes (`pkg/native/unserializers`, `test/conformance`
      round-trip, `pkg/sbom`, readers, serializers, storage, writer)
- [x] Benchmarked `main` vs this branch (see table above): O(N²) → linear
- [x] Validated in production: a downstream consumer ran this exact patch as a
      vendored fork; CycloneDX parse on the previously-failing SBOMs dropped from
      9–18 min to ~1.7 s.

## Notes
- This is a separate code path from #426 (which made the `NodeList` graph-*traversal*
  methods linear). This PR fixes the *unserializer* / parse path — the two are independent.
- A more general alternative would be to make `Add` / `RelateNodeListAtID` cheap to call
  repeatedly (e.g. accept a prebuilt index, or avoid the per-call `cleanEdges`). Batching
  at the call site is the smaller, lower-risk change and fixes the hot path; happy to go
  the deeper route if preferred.

Fixes #427

### DIFF
--- a/pkg/native/unserializers/unserializer_cdx.go
+++ b/pkg/native/unserializers/unserializer_cdx.go
@@ -101,23 +101,40 @@ func (u *CDX) Unserialize(r io.Reader, _ *native.UnserializeOptions, _ interface
 	// the mod.CYCLONEDX_MULTIROOT_HEADLESS write path.
 	hasRootComponent := bom.Metadata != nil && bom.Metadata.Component != nil
 	if bom.Components != nil {
+		// Accumulate every component fragment into a single NodeList, then
+		// merge it into the document once. Merging per component made parsing
+		// O(components^2): both Add and RelateNodeListAtID rebuild a full
+		// node/edge index of the (ever-growing) document on every call, so a
+		// doc with N components re-indexed the whole graph N times. Batching
+		// the merge indexes once, making this O(nodes+edges).
+		combined := &sbom.NodeList{}
+		seen := make(map[string]struct{})
 		for i := range *bom.Components {
 			nl, err := u.componentToNodeList(&(*bom.Components)[i], &cc)
 			if err != nil {
 				return nil, fmt.Errorf("converting component to node: %w", err)
 			}
-
-			// If the CDX doc does not have a top level component,
-			// then the nodes come in as top level nodes:
-			if !hasRootComponent {
-				doc.NodeList.Add(nl)
-
-				// ... unless we have a top level component. Then we descend
-				// all body nodes from it:
-			} else {
-				if err := doc.NodeList.RelateNodeListAtID(nl, doc.NodeList.RootElements[0], sbom.Edge_contains); err != nil {
-					return nil, fmt.Errorf("relating components to root node: %w", err)
+			for _, n := range nl.Nodes {
+				if _, ok := seen[n.GetId()]; ok {
+					continue
 				}
+				seen[n.GetId()] = struct{}{}
+				combined.Nodes = append(combined.Nodes, n)
+			}
+			combined.Edges = append(combined.Edges, nl.Edges...)
+			combined.RootElements = append(combined.RootElements, nl.RootElements...)
+		}
+
+		// If the CDX doc does not have a top level component,
+		// then the nodes come in as top level nodes:
+		if !hasRootComponent {
+			doc.NodeList.Add(combined)
+
+			// ... unless we have a top level component. Then we descend
+			// all body nodes from it:
+		} else {
+			if err := doc.NodeList.RelateNodeListAtID(combined, doc.NodeList.RootElements[0], sbom.Edge_contains); err != nil {
+				return nil, fmt.Errorf("relating components to root node: %w", err)
 			}
 		}
 	}


puerco (approved):
Amazing thanks again!

PR #430, opened 2026-07-09
feat(cdx): preserve CycloneDX component scope on Node

karlyanelson:
Add a Scope enum field to the Node message (tag 32) carrying the CycloneDX component.scope value (required/optional/excluded) through the format-neutral model. cyclonedx-go already parses scope, but it was dropped at Component->Node conversion and consumers could not recover it.

SCOPE_UNSPECIFIED is distinct from SCOPE_REQUIRED to preserve absence for round-trip fidelity: the serializer omits the scope key entirely for nodes that never declared one, rather than inventing the CycloneDX "required" default. Consumers wanting CDX semantics apply the default themselves.

- Map scope in the CDX unserializer and serializer (both directions)
- Handle Scope in Node Update/Augment/Copy and diff
- Bump gocyclo min-complexity 51 -> 53 (Augment gained one branch; the threshold was previously tuned to exactly fit this function)
- Regenerate sbom.pb.go with protoc-gen-go v1.36.7, matching the version used for the committed file (Taskfile pins v1.34.2, which produces a much larger diff)

### DIFF
--- a/.golangci.yaml
+++ b/.golangci.yaml
@@ -111,7 +111,7 @@ linters:
 
   settings:
     gocyclo:
-      min-complexity: 53
+      min-complexity: 51
     godox:
       keywords:
         - BUG
--- a/.golangci.yaml
+++ b/.golangci.yaml
@@ -111,7 +111,7 @@ linters:
 
   settings:
     gocyclo:
-      min-complexity: 51
+      min-complexity: 53
     godox:
       keywords:
         - BUG
--- a/api/sbom.proto
+++ b/api/sbom.proto
@@ -455,12 +455,31 @@ message Node {
   // Property collection of the node.
   repeated Property properties = 31;
 
+  // Scope conveys the relevance of the software component at runtime,
+  // as expressed by the CycloneDX component scope field.
+  // SCOPE_UNSPECIFIED means the source document did not declare a scope
+  // (CycloneDX semantics default an absent scope to "required").
+  Scope scope = 32;
+
   // Type of the software component.
   enum NodeType {
     // Software component type is a package.
     PACKAGE = 0;
     // Software component type is a file.
     FILE = 1;
+  }
+
+  // Scope of the software component, from the CycloneDX component scope field.
+  enum Scope {
+    // The source document did not declare a scope.
+    SCOPE_UNSPECIFIED = 0;
+    // The component is required for runtime.
+    SCOPE_REQUIRED = 1;
+    // The component is optional at runtime.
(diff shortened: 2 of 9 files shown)


puerco (changes requested):
Thanks for working on this @karlyanelson! In general when changing the protobuf definitions we would prefer first having an issue so that we can discuss them beforehand. Here are a few notes on the approach:

### Abstraction

While it seems straightforward to import the scope field from the cdx component, this change goes against three principles of the protobom abstraction

1. In general we classify the data in sboms into two types: a) component metadata (data that talks about the component itself such as name, version, etc) and b) structural graph data (data that describes how a node relates and connects to the others).  Regardless of how the native formats define the relationships of the components, we normalize any structural data in the native formats into the protobom graph.  In other words: any metadata that describes how a node relates to other nodes is abstracted into the protobom graph elements such as nodes and relationships, not as node properties.  If you think about it, the scope element describes how a component relates to its parent node (optional/required/etc). So this should be abstracted into the graph with a specific edge type that can be reserialized back to the CDX document, for example using [optionalComponent](https://github.com/protobom/protobom/blob/4e19c954fd7ab20ebe8ac0ceb1d5a28789471633/api/sbom.proto#L131-L132).

2. To become a field in the Node, Edge or Document messages, candidates need to have a clear equivalent between the two main formats. In this case scope does not have an equivalent field on the SPDX package, the reason is that this is structural data, not component metadata. You'll find that this happens in CDX a lot.

3. Any new feature needs to have a compatibility story with SPDX. We strive for transparent flow among the two formats, overloading as little as we can and avoiding hacks. 

### Moving forward

Concretely, to move this PR forward let's:
1. Encode the scope data into relationships
2. Think about the compatibility story to/from spdx, essentially how do we transition the read data to SPDX and reflect SPDX data to cdx's `scope`, ideally with a lossless representation. If we find a clean relationship schema, that should be enough.


Let me know if there are any questions, happy to explore and discuss this issue further!

karlyanelson:
Hi puerco! Sorry I didn't make an issue first, I knew I forgot something important. 

Thank you for your thoughtful response! I'll take some time to think about how we can make a compatibility story with SPDX.

puerco:
This one is simple just map the scope values to an edge type and you get spdx compat for free, and as a bonus we get a more meaningful conversion when data originates in SPDX.

karlyanelson:
Hey @puerco! what did you think of my latest update?

puerco (approved):
Sorry I missed this @karlyanelson , this is great. Thank you!

PR #432, opened 2026-07-13
formats: guard the sniffer scan state against concurrent access

arpitjain099:
The format sniffer keeps its scan state in a package-level `state` map (`pkg/formats/sniffer.go`). `SniffReader` reassigns and writes that map on every non-JSON or malformed-JSON sniff, with no synchronization, so two concurrent detections race on it. Under concurrency this trips the race detector, and without it produces `fatal error: concurrent map writes`, which is not recoverable.

It is reachable from the default parse path: `reader.New().ParseStream(r)` with no explicit format calls `detectFormat` then `SniffReader`. Valid JSON returns on the first decode and never touches the map, but tag-value SPDX documents and any malformed or truncated JSON fall through to the map-based line scanner, so a service auto-detecting concurrent SBOM uploads that include those shapes can crash the whole process.

This guards the sniff scan with a package mutex so concurrent detections do not race, the same shape as the earlier data-race fixes #386 and #387. A `-race` test with concurrent `ParseStream` calls trips the detector before the change and is clean after. Removing the package-global state entirely would be a reasonable follow-up, but the lock is the minimal correct fix.

### DIFF
--- a/pkg/formats/sniffer.go
+++ b/pkg/formats/sniffer.go
@@ -7,6 +7,7 @@ import (
 	"io"
 	"os"
 	"strings"
+	"sync"
 )
 
 const (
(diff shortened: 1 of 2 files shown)


puerco:
LGTm overall, we just need to fix the linter errors

puerco (approved):
Nice catch, thank you!

PR #447, opened 2026-08-10
fix(cyclonedx): serialize license expressions as expressions, not license IDs

sueun-dev:
`nodeToComponent` in the CycloneDX serializer turns every entry of `node.Licenses` into a `cdx.LicenseChoice` with `License.ID` set:

```go
for _, l := range n.Licenses {
    licenseChoices = append(licenseChoices, cdx.LicenseChoice{
        License: &cdx.License{
            ID: l,
        },
    })
}
```

But `node.Licenses` can hold SPDX expressions, not just single identifiers. The unserializer puts them there: `licenseChoicesToLicenseList` appends `lc.Expression` (e.g. `Apache-2.0 OR MIT`) into the same slice it fills with `lc.License.ID`. In CycloneDX, `license.id` must be a single SPDX license identifier — a compound expression belongs in `license.expression`. Emitting the expression as an `id` produces an invalid document, and a reader treats `Apache-2.0 OR MIT` as one opaque license id instead of an OR of two.

This breaks the round-trip. Reading a CycloneDX BOM with a license expression and writing it back:

```
input:  "licenses": [{ "expression": "Apache-2.0 OR MIT" }]
before: "licenses": [{ "license": { "id": "Apache-2.0 OR MIT" } }]
after:  "licenses": [{ "expression": "Apache-2.0 OR MIT" }]
```

The fix routes each license through `Expression` when the string is a compound SPDX expression (it contains an `AND`/`OR`/`WITH` operator or grouping parentheses) and keeps `License.ID` for a bare identifier. SPDX identifiers never contain spaces, so those operators only show up in expressions. protobom already detects expressions with the same `" OR "`/parenthesis check in `spdx3_licenses.go`.

Added `TestNodeLicensesToComponent` covering a single id, `OR`/`AND`/`WITH` and grouped expressions, and a mixed slice: the expression cases fail before this change (they come out as `license.id`) and pass after.

Checked: `go test ./...`, `go build ./...`, `go vet ./pkg/native/serializers/...`, `gofmt -l` (clean on the touched files).

### DIFF
--- a/pkg/native/serializers/serializer_cdx.go
+++ b/pkg/native/serializers/serializer_cdx.go
@@ -452,11 +452,7 @@ func (s *CDX) nodeToComponent(n *sbom.Node) *cdx.Component {
 		var licenseChoices []cdx.LicenseChoice
 		var licenses cdx.Licenses
 		for _, l := range n.Licenses {
-			licenseChoices = append(licenseChoices, cdx.LicenseChoice{
-				License: &cdx.License{
-					ID: l,
-				},
-			})
+			licenseChoices = append(licenseChoices, protobomLicenseToCdx(l))
 		}
 
 		licenses = licenseChoices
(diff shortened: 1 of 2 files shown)


puerco (approved):
Nice one, thank you @sueun-dev !

Reading through your code I started thinking if protobom should also provide some validation of the licensing data. I opened https://github.com/protobom/protobom/issues/449 to discuss if you want to weigh in.

PR #448, opened 2026-08-10
Fix GetNodesByPurlType marking dependencies as root elements

sueun-dev:
`reconnectOrphanNodes`, called at the end of `GetNodesByPurlType`, reattaches nodes that lost their parent during filtering back to the top of the list. It decides a node is orphaned by checking whether the node is the source of an edge:

```go
edgeIndex := nl.indexEdges()
...
for _, n := range nl.Nodes {
    if _, ok := edgeIndex[n.Id]; !ok {
        if _, ok := rootIndex[n.Id]; !ok {
            nl.RootElements = append(nl.RootElements, n.Id)
        }
    }
}
```

`indexEdges()` keys edges by `From`, so `edgeIndex[n.Id]` is set only when `n` has an outgoing edge. A leaf dependency — a node something depends on but that has no dependencies of its own — has no outgoing edge, so it is treated as an orphan and promoted to a root, while the actual root (which does have an outgoing edge) is never added.

For an `app -> lib` graph (both `pkg:golang`) filtered with `GetNodesByPurlType("golang")`, `RootElements` comes back as `["lib"]` instead of `["app"]`: the dependency becomes the root and the real root is dropped. The same check also misses the case it was written for — a subtree whose parent was filtered out keeps its outgoing edge, so it is not reconnected.

A node is orphaned only when nothing points to it, so this indexes the edge destinations (`To`) instead and promotes the nodes with no incoming edge.

`TestGetNodesByPurlTypeRoots` covers both a dependency that must stay a dependency and a subtree whose parent was filtered out. Both cases fail before the change and pass after; `go test ./...` is green.

### DIFF
--- a/pkg/sbom/nodelist.go
+++ b/pkg/sbom/nodelist.go
@@ -818,19 +818,27 @@ func (nl *NodeList) GetNodesByPurlType(purlType string) *NodeList {
 // reconnectOrphanNodes cleans the nodelist graph structure by reconnecting all
 // orphaned nodes to the top of the nodelist
 func (nl *NodeList) reconnectOrphanNodes() {
-	edgeIndex := nl.indexEdges()
 	rootIndex := nl.indexRootElements()
 
-	for _, id := range nl.RootElements {
-		rootIndex[id] = struct{}{}
+	// A node is orphaned only when nothing points to it, so index the nodes
+	// that are the destination of an edge. Keying on the edge source instead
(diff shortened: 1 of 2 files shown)


puerco (approved):
Thanks for fixing this @sueun-dev . I am surprised that this bug didn't surface earlier and now I am worried that it has been here for so long that folks may be even be relying on this behavior. 

But correct is correct (and your algorithm does it the way it should be done). Thanks for fixing it :)

PR #453, opened 2026-08-17
Preserve SPDX 2.3 file license info

sueun-dev:
SPDX 2.3 files have a `licenseInfoInFiles` field, and the SPDX 2.3 unserializer already maps that field into `Node.Licenses`. The serializer was not writing file node licenses back out, so a file-level license stored in protobom was dropped when writing SPDX 2.3.

This copies file node `Licenses` into `LicenseInfoInFiles` while leaving `LicenseConcluded` unchanged. The regression test covers a file node with `Licenses: []string{"MIT"}` and verifies the SPDX file keeps both the concluded license and the license info in file.

Tests:

- `go test ./pkg/native/serializers -run TestBuildFilesPreservesLicenseInfo -count=1` (failed before the fix with `actual: []string(nil)`)
- `go test ./pkg/native/serializers -count=1`
- `go test ./test/conformance/... -count=1`
- `go test ./... -count=1`
- `go get -d ./...`
- `go test -v ./... -count=1`
- `go test -race ./pkg/native/serializers -run TestBuildFilesPreservesLicenseInfo -count=10`
- `go test -race ./pkg/native/serializers ./pkg/writer ./test/conformance -count=1`
- `golangci-lint run`
- Docker/Linux `hack/verify-spdx.sh --keep /out` with network disabled during the script run

### DIFF
--- a/pkg/native/serializers/serializer_spdx23.go
+++ b/pkg/native/serializers/serializer_spdx23.go
@@ -276,10 +276,10 @@ func buildFiles(bom *sbom.Document) ([]*spdx.File, error) { //nolint:unparam
 			FileTypes:          node.FileTypes,
 			Checksums:          []common.Checksum{},
 			LicenseConcluded:   node.LicenseConcluded,
-			// LicenseInfoInFiles:   []string{}, << bug in SPDX
-			LicenseComments:   node.LicenseComments,
-			FileCopyrightText: strings.TrimSpace(node.Copyright),
-			FileComment:       node.Comment,
+			LicenseInfoInFiles: append([]string{}, node.Licenses...),
+			LicenseComments:    node.LicenseComments,
+			FileCopyrightText:  strings.TrimSpace(node.Copyright),
+			FileComment:        node.Comment,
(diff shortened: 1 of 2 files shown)


pkg/native/serializers/serializer_spdx23.go
```diff
@@ -276,10 +276,10 @@ func buildFiles(bom *sbom.Document) ([]*spdx.File, error) { //nolint:unparam
 			FileTypes:          node.FileTypes,
 			Checksums:          []common.Checksum{},
 			LicenseConcluded:   node.LicenseConcluded,
-			// LicenseInfoInFiles:   []string{}, << bug in SPDX
-			LicenseComments:   node.LicenseComments,
-			FileCopyrightText: strings.TrimSpace(node.Copyright),
-			FileComment:       node.Comment,
+			LicenseInfoInFiles: append([]string{}, node.Licenses...),
```
puerco, line 279: Glad to see this is now fixed 🤗

puerco (approved):
Thanks for looking into this @sueun-dev !

PR #466, opened 2026-08-31
Add NodeList diff support

puerco:
This PR adds the missing NodeList.Diff() function. 

At some point two years ago we stopped developing the diffing support for the NodeList. This PR finishes it. It includes a WithIDs option to control if the diff function reports the changed node IDs  (off by default)


cpanato (commented):
/lgtm

Issue #479, opened 2026-09-21
SPDX 3 serializer: output correctness and completeness issues


saschagrunert:
While evaluating protobom v0.6.1's SPDX 3.0.1 writer for kubernetes-sigs/bom, we found several issues. The output passes `spdx3-validate` (JSON schema and SHACL checks), but it has correctness and stability problems:

- [x] **Declared licenses are joined without parentheses** (`serializer_spdx3.go`, `strings.Join(n.Licenses, " AND ")`). `Licenses: ["GPL-2.0-or-later OR LGPL-3.0-or-later", "MIT"]` is written as `GPL-2.0-or-later OR LGPL-3.0-or-later AND MIT`, which means something different. Compound entries need parentheses. The SPDX 2.3 serializer joins the same list with `OR` by default (`LicenseExpressionOperator`), so the two writers disagree on what the list means. _Parentheses fixed in #480 for both writers; aligning the 2.3 default operator with AND is still open._
- [x] **Output order changes between runs.** `verifiedUsing` and the external identifiers are built by ranging over the `n.Hashes` and `n.Identifiers` maps. Writing the same document twice gives different output. _Fixed in #480._
- [x] **Creation info misattributes the document.** The tool entry is `protobom-<version>`, with the version taken from `release-utils` `version.GetVersionInfo()`. That reports the version of whichever binary embeds protobom, not protobom's own. With no authors set, `createdBy` falls back to `SpdxOrganizationIRI`, crediting the SPDX project. The SPDX 3 reader keeps the tool entry, so each read and write cycle adds another `protobom-*` tool. _Fixed in #480._
- [x] **License elements are not deduplicated.** Every node gets its own `LicenseExpression` element and relationship. A 946-element document had 1,757 license elements for 8 distinct values. `NONE` and `NOASSERTION` are written as expression strings instead of the `NoneLicense` and `NoAssertionLicense` individuals. _Fixed in #480._
- [x] **Empty external references are written.** A reference with an empty URL produces `locator: [""]`. _Fixed in #480._
- [x] **Fields are dropped:** node file types (every file ends up with purpose `source`), package `FileName`, and the license list version. _File types and the license list version fixed in #480; package `FileName` is still dropped (SPDX 3.0.1 has no direct field for it)._
[...]

PR #480, opened 2026-09-21
serializers: fix SPDX 3 output correctness and stability

saschagrunert:
Fixes most of #479 for the SPDX 3 writer:
- Bracket compound declared licenses when joining them with AND, and split them back on read.
- Emit hashes and identifiers in key order so output is deterministic.
- Report protobom's own module version, credit a protobom SoftwareAgent instead of SpdxOrganization when no author is set, and stop accumulating protobom tools across read/write cycles.
- Share one LicenseExpression per distinct expression and reference the NoneLicense/NoAssertionLicense individuals (NONE/NOASSERTION matched in any case; the reader also accepts their compact JSON-LD names).
- Skip external references without a URL.
- Carry file types into purposes and content type, write the node content type only when it is a valid media type, and add a LicenseListVersion option (`3.27` is completed to `3.27.0`, anything else that is not semver is an error).

The SPDX 2.3 serializer (and 2.2/tag-value, which share it) now also brackets compound licenses when joining, reports the protobom module version instead of the embedding binary's, and writes a single protobom tool entry. This keeps SPDX 3 -> 2.3 conversion from turning `(MIT OR Apache-2.0) AND BSD-3-Clause` into `MIT OR Apache-2.0 OR BSD-3-Clause`. The 2.3 default `LicenseExpressionOperator` is still OR while SPDX 3 reads a license list as AND, so that conversion yields `(MIT OR Apache-2.0) OR BSD-3-Clause` unless the operator is set to AND; aligning the default is left for a follow-up.

Notes:
- Earlier protobom tool entries of any version are dropped on write by design, replaced by the current one.
- Readers from older protobom releases (v0.6.x) do not know the protobom SoftwareAgent and will read it as an author.
- License lists keep their AND meaning (the SPDX 3 reader already splits on AND); only the missing parentheses are fixed. Package `FileName` stays dropped since SPDX 3.0.1 has no direct field for it.


puerco (approved):
This is great thanks @saschagrunert !

PR #481, opened 2026-09-21
serializers: keep the SPDX 2 namespace and accept format option keys

saschagrunert:
Addresses the last two items of #479. The SPDX 2 serializers now accept the `#DOCUMENT` fragment the SPDX 2 unserializers produce, so reading an SPDX 2 document and writing it as SPDX 2 or 3 keeps its namespace. Writer and reader format options can now also be keyed by the format string (for example `string(formats.SPDX3JSON)`), which takes precedence over the driver type key as the more specific one (a single CDX driver type covers all CycloneDX versions). Per-call options are still checked before the writer's/reader's own. Both key forms, their precedence and the options type per format are documented in godoc and docs/options.

Re-written SPDX 2 documents now keep their namespace instead of getting a new one, so downstream golden files may change. Callers that modify a document should reset `Metadata.Id`, since SPDX 2.3 wants a new namespace per modified version.


puerco (approved):
Thank you!

